PR #3, opened 2017-01-06
Add Possible Time Save Component

hallcristobal:


src/component/mod.rs
```diff
@@ -2,4 +2,5 @@ pub mod timer;
 pub mod title;
 pub mod splits;
 pub mod previous_segment;
-pub mod sum_of_best;
\ No newline at end of file
+pub mod sum_of_best;
+pub mod possible_time_save;
```
CryZe: Can you add this to your VS Code settings?

```json
"files.insertFinalNewline": true,
```

hallcristobal: Sure no problem.

PR #6, opened 2017-01-14
Correct formatting of Possible Time Save Component

hallcristobal:
Corrected formatting of Possible Time Save Component

### DIFF
--- a/src/time_formatter/possible_time_save.rs
+++ b/src/time_formatter/possible_time_save.rs
@@ -23,7 +23,7 @@ impl PossibleTimeSave {
 
 impl Default for PossibleTimeSave {
     fn default() -> Self {
-        PossibleTimeSave { accuracy: Accuracy::Tenths }
+        PossibleTimeSave { accuracy: Accuracy::Hundredths }
     }
 }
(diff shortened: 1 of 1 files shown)


src/segment.rs
```diff
@@ -59,6 +60,7 @@ impl Segment {
 
     #[inline]
     pub fn set_best_segment_time(&mut self, time: Time) {
+        self.comparisons.insert(BEST_SEGMENT_COMPARISON_NAME.into(), time);
```
CryZe: That's not what the C# code does. What's the purpose of this?

hallcristobal: The live split code adds the best segment to the comparisons via the best segments comparison generator, however that's not implemented yet but you are indexing comparisons on best segments which is returning None causing the live delta to not work correctly

CryZe: well, I'd rather have it not working fully atm, than introducing code that makes it temporarily work and cause bugs later on if we forget to remove it.

hallcristobal: that's fair enough, I'll remove it

PR #25, opened 2017-04-15
Implement a Timer History

CryZe:
This is still very much work in progress. This probably needs some more / better thought out CoW layers. Also the history is unbounded at the moment. There should be a configurable bound and you should also be able to turn it off completely.

This is also blocked by the next Rust release, as we need to do ptr_eq checks on Arc: https://doc.rust-lang.org/std/sync/struct.Arc.html#method.ptr_eq


CryZe:
Considering it's completely unclear whether we want this, it's super outdated with tons of conflicts and I don't intend on working on it anytime soon, since I'd rather work on subsplits, auto splitting, rendering, UI and co., I'm going to close this. We can look into this eventually again.

PR #27, opened 2017-05-24
Implement Auto Splitters

CryZe:
This implements Auto Splitters. This is still a huge Work in Progress as it's still entirely unclear which scripting language to use. These are the options:

- Lua: This is probably one of the most viable options. Unfortunately it adds a C Dependency which can cause compilation / linking issues. I can't compile it on my Windows machine apparently, at least with the GNU Toolchain, which I definitely need to fix, in order to choose Lua. However I might also be able to just use the MSVC Toolchain from now on with livesplit-core. If that works, we probably want Lua. The bindings I would use would either be hlua, which I've used before, or the new rlua bindings.

- Dyon: This seems to be the most viable scripting language written in Rust. It should be able to compile anywhere, but its target audience is game development, so it has a very specific style to it and doesn't have any 64-bit integers or hexadecimal literals, which is extremely bad for our purpose. Also it doesn't come with any standard library, so we would need to provide functions for basic things like working with strings, files, network, ...

- Rhai: Yeah no. This one doesn't seem to have progressed enough to even be usable in any usecase for anybody, so this seems very much not like an option.

- gluon: This is written in Rust and comes with its own standard library. The only problem is that its heavily functional, which might be too hard for most people writing the ASL scripts, so I'm not too sure about this one.

- ketos: Seems kind of nice, but is a Lisp Dialect, so it looks pretty ugly to use. I'd prefer gluon's functional style over this.

---

Operating System Support:

- [x] Windows
- [ ] Linux
- [ ] macOS


CryZe:
I should consider building it on top of this library: https://github.com/luser/read-process-memory

CryZe:
I should consider building it on top of this library: https://github.com/luser/read-process-memory

CryZe:
Wren may also be a good scripting language to use.

CryZe:
Considering we are going for a totally different approach now, I'm going to close this.

PR #62, opened 2017-07-14
Add Possible Time Save to Previous Segment

hallcristobal:


src/component/previous_segment.rs
```diff
@@ -152,12 +170,20 @@ impl Component {
...
         let text = self.text(live_segment, resolved_comparison);
+        let mut time = Delta::custom(self.settings.drop_decimals, self.settings.accuracy)
+            .format(time_change)
+            .to_string();
+
+        if self.settings.show_possible_time_save {
+            time.push_str(" / ");
+            time.push_str(&PossibleTimeSave::with_accuracy(self.settings.accuracy)
+                .format(previous_possible)
+                .to_string());
```
CryZe: That allocates a string, pushes it onto the other and then deallocates the allocated string, which is wasteful. You can use `write!(time, " / {}", PossibleTimeSave::with...).unwrap()` to directly format into the time String. The unwrap is fine as that can never fail (but it could for other types that are not Strings)

PR #67, opened 2017-08-18
For basic getters and setters, use public fields instead

mark-blum:
A few structs (`Run`, `RunMetadata`, `Segment`) used a lot of getter's and setter's which didn't add much, if anything at all, to just having that field be public. Those kinds empty getters and setters are considered bad practice as they clutter up the api unnecessarily, and this pull request adresses that.

The main files to look at here are `src/run/run.rs`, `src/run/segment.rs`, and `src/run/run_metadata.rs`, as they contain the api-changing sections of code. The rest of the files are mainly cascading the api changes to the rest of this crate. The philosophy I took for what to change to public fields was based on if the field had both a getter and a setter/mut function and both of these functions boiled down to a `&self.field` and `self.field = input` (or `&mut self.field`).

If you have any questions about the changes, please let me know.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@ emscripten/target
 emscripten/Cargo.lock
 js/livesplit.js
 Cargo.lock
-**.rs.bk
+*.rs.bk
+*.swp
--- a/capi/src/run.rs
+++ b/capi/src/run.rs
@@ -120,7 +120,7 @@ pub unsafe extern "C" fn Run_len(this: *const Run) -> usize {
 
 #[no_mangle]
 pub unsafe extern "C" fn Run_segment(this: *const Run, index: usize) -> *const Segment {
-    acc(this).segments.get(index).unwrap_or(std::ptr::null())
+    acc(this).segments.get(index).map_or(ptr::null(), |p| p as *const Segment)
 }
 
 #[no_mangle]
--- a/capi/src/run.rs
+++ b/capi/src/run.rs
@@ -38,32 +38,36 @@ pub unsafe extern "C" fn Run_clone(this: *const Run) -> OwnedRun {
 
 #[no_mangle]
 pub unsafe extern "C" fn Run_push_segment(this: *mut Run, segment: OwnedSegment) {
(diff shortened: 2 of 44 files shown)


CryZe:
I'm not sure if I agree with this. As you saw, I wrote ton a whole lot of these accessors, so I definitely must've thought about this as well, cause typing those down can't have been fun. But, I chose to do it anyway, because having a more stable API outweighs choosing the more lazy approach in the long run. If you take a look at the Rust standard library, the same approach is chosen there as well. I don't think there's any struct with public fields in the standard library. There is a place for those, especially when you have "JSON like" / Plain old data like structs, that are just some information that you pass around, like the state and configuration objects in livesplit-core. But once you have some Object (in the actual OOP sense) then you need to ensure the consistency of your data and accessors are the right approach, even when slightly annoying to type down. However there is also the argument of that causing some more borrowing issues, but usually that just means you are reaching too far into that object anyway, and that object should have some entry / retain like API for modifying it in this special way that is currently troubling you. So yeah, I don't think this change is something that I would want. I really appreciate that you are trying to help and that you bring up these potential flaws in the code, but unfortunately I don't think I want to merge this.

mark-blum:
The difference between this and the standard library is that most of the code in the standard library is actually abstracting something that couldn't just have a public field. Take for instance `Vec::get()`. This does bound checks and does the necessary unsafe pointer dereferencing to return the proper result. The standard library isn't generally a good comparison in this case however since most of their code is not just holding "plain data" as we are in this case.

But regardless what I'm trying to say is that there is nothing wrong with having a public field, and it certainly seems a whole lot better than having a bunch of `best_segment_time()`, `best_segment_time_mut()`, `set_best_segment_time(...)` that all do exactly what

CryZe:
These things might do something in the future though. Setting the best segment time for example might mean that the segment time or some history might need to be recalculated or something. So it's not a good idea for stability to have them as public fields. When there is certain constraints on the validity of an object, then the fields really shouldn't be public. I'm fine with the "ugliness" of the API. Thanks for trying to help though.

PR #71, opened 2017-09-06
Use actual lifetimes instead of 'static for the `acc`, `acc_mut` and `str` functions

mark-blum:
The current capi has convenience functions for converting raw pointers into references or `str`s, however it gives them `'static` lifetimes which is quite unsafe as it practically disables many lifetime checks that would normally happen with the those references. This PR remedies this issue.

The main changes are in `capi/src/lib.rs` and the rest of the changes are just propagating those changes.

There are a few corner cases that these enhanced functions can't handle, particularly certain iterators such as `OwnedRunMetadataVariablesIter`, since raw pointers can't have lifetime information, however because of the enhanced functions we can easily see these situations and document their lifetime dependence, as I have done in those cases.

### DIFF
--- a/capi/src/run_metadata.rs
+++ b/capi/src/run_metadata.rs
@@ -25,9 +25,11 @@ pub unsafe extern "C" fn RunMetadata_region_name(this: *const RunMetadata) -> *c
     output_str(acc(&this).region_name())
 }
 
+/// # Safety
+/// `this` must outlive `OwnedRunMetadataVariablesIter`
 #[no_mangle]
-pub unsafe extern "C" fn RunMetadata_variables(
+pub unsafe extern "C" fn RunMetadata_variables<'a>(
     this: *const RunMetadata,
-) -> OwnedRunMetadataVariablesIter {
-    alloc(acc(&this).variables())
+) -> OwnedRunMetadataVariablesIter<'a> {
+    alloc((&*this).variables())
 }
--- a/capi/src/run_metadata.rs
+++ b/capi/src/run_metadata.rs
@@ -7,27 +7,27 @@ pub type OwnedRunMetadata = *mut RunMetadata;
 
 #[no_mangle]
 pub unsafe extern "C" fn RunMetadata_run_id(this: *const RunMetadata) -> *const c_char {
-    output_str(acc(this).run_id())
+    output_str(acc(&this).run_id())
 }
 
 #[no_mangle]
 pub unsafe extern "C" fn RunMetadata_platform_name(this: *const RunMetadata) -> *const c_char {
-    output_str(acc(this).platform_name())
+    output_str(acc(&this).platform_name())
 }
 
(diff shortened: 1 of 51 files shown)


CryZe:
Can you show me some example where this actual benefits me in any way? All of this is only called by C (and the other languages), and for none of those these lifetimes have any benefit. It might make the implementation slightly more correct, but those are mostly just one liners, so there's not really any benefit there either. I might merge this, as it isn't that bad (and I might use these lifetimes for eventually carrying the lifetimes over to JavaScript and co. which then can maybe do borrow checking at runtime). But in Rust itself I see almost no benefit (but maybe there is).

CryZe:
I like the safety comments though. I might be able to use the for the binding generator to carry them over to all the bindings.

mark-blum:
It's true since this is just in the C Api it doesn't have a huge impact, as most of the time we only need the pointer to live as long as the function - which is mostly guaranteed - but to me the biggest value this adds is in those cases where we need the `this` pointer to live longer than what we return, as is the case with the `OwnedRunMetadataVariablesIter`. In those cases the compiler won't let you use the convenience function anymore since we can't pass along that lifetime information, which means we must manually cast, which is a sign that we should document the special lifetime conditions of this function.

Additionally, it prevents us from accidentally doing silly things on the Rust side, like putting it into some `static Vec<&'static SomeStruct>`, even if it is very unlikely in this case.

capi/src/run_metadata.rs
```diff
@@ -7,27 +7,29 @@ pub type OwnedRunMetadata = *mut RunMetadata;
...
 
+/// # Safety
+/// `this` must outlive `OwnedRunMetadataVariablesIter`
 #[no_mangle]
-pub unsafe extern "C" fn RunMetadata_variables(
+pub unsafe extern "C" fn RunMetadata_variables<'a>(
     this: *const RunMetadata,
-) -> OwnedRunMetadataVariablesIter {
-    alloc(acc(this).variables())
+) -> OwnedRunMetadataVariablesIter<'a> {
```
CryZe: This has no effect whatsoever though. A lifetime always needs to be bound to at least two "things", otherwise the compiler treats it as 'static. So even in this case you don't gain anything from it.

CryZe: However I agree that documenting it is very useful. So I would definitely merge something that just adds some documentation (or maybe the lifetime in these and only these cases).

mark-blum: You're right that it doesn't really have an effect, since raw pointers don't carry lifetime information, however this is also about the intent. If someone new looked at this library and saw that lifetime operator, they would be able to see that it must be dependent on some other object, even if the compiler can't enforce that here.

The old `'static` annotations on make me think that some it must reference some static object, which is wrong in this case as it's just a little hack to avoid a couple of lifetime annotations.

PR #73, opened 2017-09-06
Remove Timer.{replace, set}_run()

mark-blum:
I'm finding it very hard to think of a situation where this is actually
useful, safe for trying to save an allocation, but even then the
allocation is quite small and it is usually better to just create a new
Timer.


CryZe:
You want to be able to switch between different runs while keeping your timer's current configuration (comparison, game time / real time, ...).

These are useful methods.

CryZe:
I'm closing these since, replace_run does some additional logic that is important for ensuring the correctness of a Timer object.

PR #74, opened 2017-09-08
Change Timer.current_split_index from isize to Option<usize>

mark-blum:
It is the more rustic way and surprisingly doesn't effect the C api as far as I can tell. Also any panics from the unwraps (which should only be in places where there was some prior check of `TimerPhase::Running` or otherwise) would let us know of some logic error instead of just silently casting a `-1` isize to usize.


CryZe:
Alright, this looked good, so I rebased your changes and merged them in https://github.com/CryZe/livesplit-core/commit/50d263d0028fffaa10c1256053b50747e0482b5f

PR #81, opened 2017-11-04
Close Multiple Issues

### LINKED ISSUE #70: Modify CheckLiveDelta to return a segment delta when necessary
This simplifies the logic in Previous Segment slightly. (It also affects the segment delta columns in Splits and Subsplits, but those aren't implement in ls-core yet.)

https://github.com/LiveSplit/LiveSplit/commit/1e258ab5ae530fd409b54865889b854f11eddd76

hallcristobal:
Closes #70

### DIFF
--- a/capi/src/setting_value.rs
+++ b/capi/src/setting_value.rs
@@ -1,4 +1,4 @@
-use livesplit_core::settings::{Color, Gradient, Value as SettingValue};
+use livesplit_core::settings::{Alignment, Color, Gradient, Value as SettingValue};
 use livesplit_core::time::formatter::{Accuracy, DigitsFormat};
 use livesplit_core::TimingMethod;
 use {alloc, own_drop, str};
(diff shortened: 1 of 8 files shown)


CryZe:
Why did you close this?

hallcristobal:
Had more on local that needed to be commited

CryZe:
So are you reopening it at some point?

hallcristobal:
in like 2 minutes lol

src/run/parser/composite.rs
```diff
@@ -23,6 +23,8 @@ pub struct ParsedRun {
 }
 
 fn parsed(run: Run, kind: TimerKind) -> ParsedRun {
+    let mut run = run;
+    run.fix_splits();
```
CryZe: I wonder if this is the appropriate way to fix the splits when "opening them". Maybe that's something LiveSplit One should do instead, or maybe whenever the Timer sets a Run? (cc @wooferzfg)

hallcristobal: I was thinking that or possibly when they're done being parsed, in the actual parse function, however that would have the same result as this in the end. But having it in livesplit-core would solidify that they would be fixed regardless of implementation.

wooferzfg: Yeah, I think it makes more sense for LSO to do it. Also, run.fix_splits() needs to be called before saving splits as well; I'm not sure if we do that already or not.

wooferzfg: If livesplit-core handled fixing splits when parsing and before saving that might work well too.

CryZe:
I think I want to experiment with fixing in LSO and then make that decision. So I think it would be best if you'd split that out so I can merge the rest.

hallcristobal:
Updated to remove Fixing

PR #85, opened 2017-11-29
Introduce catch blocks to clean up the code

### LINKED ISSUE #84: Clean Up Option Usage
The question mark operator for Option just landed, so we can simplify a lot of our Option usage, especially around optional TimeSpans. The question mark operator however fully returns out of the current function, so you'd need the unstable catch blocks in a lot of cases. However a simple catch macro, makes up for it:

```rust
macro_rules! catch {
    ($($rest:tt)*) => {
        (|| { $($rest)* })()
    }
}

fn main() {
    let a = Some(5);
    let b = Some(3);
    
    let c = catch! {
        Some(2 * a? + 5 * b?)
    };
    
    println!("{:?}", c);
}
```

CryZe:
With the recent introduction of the question mark operator for Options, we can now clean up the codebase a lot by using small catch blocks, instead of using option_op, map, and_then, if let, etc. all over the place.

Closes #84

### DIFF
--- a/src/analysis/current_pace.rs
+++ b/src/analysis/current_pace.rs
@@ -14,18 +14,18 @@ pub fn calculate(timer: &Timer, comparison: &str) -> Option<TimeSpan> {
                 timing_method,
             ).unwrap_or_default();
 
-            let live_delta = TimeSpan::option_sub(
-                timer.current_time()[timing_method],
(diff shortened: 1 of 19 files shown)


CryZe:
@wooferzfg Can you check that I didn't break anything?

PR #88, opened 2017-12-01
Modify add_custom_comparison to return Result

hallcristobal:
Changed add_custom_comparison to return a result to provide feedback for
whether or not it added it successfully, as well as incorperated name and
duplicate checking into the function instead of in a seperate method.

### DIFF
--- a/src/run/editor/mod.rs
+++ b/src/run/editor/mod.rs
@@ -7,7 +7,7 @@
 use std::num::ParseIntError;
 use std::mem::swap;
 use {comparison, unicase, Image, Run, Segment, Time, TimeSpan, TimingMethod};
-use super::run::ComparisonError;
+use super::run::{ComparisonError, ComparisonResult};
 use time::ParseError as ParseTimeSpanError;
 
 pub mod cleaning;
@@ -40,6 +40,19 @@ quick_error! {
         /// The Run Editor couldn't be opened because an empty run with no
         /// segments was provided.
         EmptyRun {}
+    }
+}
+
+quick_error! {
+    /// Error type for a failed Rename
+    #[derive(PartialEq, Debug)]
+    pub enum RenameError {
+        /// Old Comparison was not found durring rename.
+        OldNameNotFound {}
+        /// Name was invalid
+        InvalidName(err: ComparisonError) {
+            from()
+        }
     }
 }
 
@@ -572,10 +585,7 @@ impl Editor {
 
     /// Adds a new custom comparison. It can't be added if it starts with
     /// `[Race]` or already exists.
-    pub fn add_comparison<S: Into<String>>(
-        &mut self,
-        comparison: S,
-    ) -> Result<(), ComparisonError> {
+    pub fn add_comparison<S: Into<String>>(&mut self, comparison: S) -> ComparisonResult<()> {
         let comparison = comparison.into();
         self.run.add_custom_comparison(comparison)?;
         self.fix();
@@ -589,7 +599,7 @@ impl Editor {
         &mut self,
         run: &Run,
         comparison: S,
-    ) -> Result<(), ComparisonError> {
+    ) -> ComparisonResult<()> {
         let comparison = comparison.into();
         self.run.add_custom_comparison(comparison.as_str())?;
         // TODO Borrowcheck (remaining_segments isn't used after this block
@@ -651,33 +661,32 @@ impl Editor {
 
     /// Renames a comparison. The comparison can't be renamed if the new name of
     /// the comparison starts with `[Race]` or it already exists.
-    pub fn rename_comparison(&mut self, old: &str, new: &str) -> Result<(), ()> {
+    pub fn rename_comparison(&mut self, old: &str, new: &str) -> Result<(), RenameError> {
         if old == new {
             return Ok(());
         }
 
-        match self.run.validate_comparison_name(new) {
-            Ok(()) => {
-                let position = self.run
-                    .custom_comparisons()
-                    .iter()
-                    .position(|c| c == old)
-                    .ok_or(())?;
+        self.run.validate_comparison_name(new)?;
 
-                self.run.custom_comparisons_mut()[position] = new.to_string();
+        {
+            let comparison_name = self.run.custom_comparisons_mut()
+                .iter_mut()
+                .find(|c| *c == old)
+                .ok_or(RenameError::OldNameNotFound)?;
 
-                for segment in self.run.segments_mut() {
-                    if let Some(time) = segment.comparisons_mut().remove(old) {
-                        *segment.comparison_mut(new) = time;
-                    }
-                }
-
-                self.fix();
-
-                Ok(())
-            }
-            _ => Err(()),
+            comparison_name.clear();
+            comparison_name.push_str(new);
         }
+
+        for segment in self.run.segments_mut() {
+            if let Some(time) = segment.comparisons_mut().remove(old) {
+                *segment.comparison_mut(new) = time;
+            }
+        }
+
+        self.fix();
+
+        Ok(())
     }
 
     /// Clears out the Attempt History and the Segment Histories of all the
--- a/src/run/editor/mod.rs
+++ b/src/run/editor/mod.rs
@@ -7,7 +7,7 @@
 use std::num::ParseIntError;
 use std::mem::swap;
 use {comparison, unicase, Image, Run, Segment, Time, TimeSpan, TimingMethod};
-use super::validate_comparison_name;
+use super::run::ComparisonError;
 use time::ParseError as ParseTimeSpanError;
 
 pub mod cleaning;
@@ -572,7 +572,10 @@ impl Editor {
 
     /// Adds a new custom comparison. It can't be added if it starts with
     /// `[Race]` or already exists.
-    pub fn add_comparison<S: Into<String>>(&mut self, comparison: S) -> Result<(), ()> {
+    pub fn add_comparison<S: Into<String>>(
+        &mut self,
+        comparison: S,
+    ) -> Result<(), ComparisonError> {
         let comparison = comparison.into();
         self.run.add_custom_comparison(comparison)?;
         self.fix();
@@ -586,7 +589,7 @@ impl Editor {
         &mut self,
         run: &Run,
         comparison: S,
-    ) -> Result<(), ()> {
+    ) -> Result<(), ComparisonError> {
         let comparison = comparison.into();
         self.run.add_custom_comparison(comparison.as_str())?;
         // TODO Borrowcheck (remaining_segments isn't used after this block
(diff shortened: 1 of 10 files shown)


src/run/editor/mod.rs
```diff
@@ -465,8 +465,7 @@ impl Editor {
 
     pub fn add_comparison<S: Into<String>>(&mut self, comparison: S) -> Result<(), ()> {
         let comparison = comparison.into();
-        if validate_comparison_name(&self.run, &comparison) {
-            self.run.add_custom_comparison(comparison);
+        if self.run.add_custom_comparison(comparison).is_ok() {
```
CryZe: You can use the ? operator to propagate the error up instead of using if ....().is_ok()

src/run/editor/mod.rs
```diff
@@ -480,9 +479,7 @@ impl Editor {
         comparison: S,
     ) -> Result<(), ()> {
         let comparison = comparison.into();
-        if validate_comparison_name(&self.run, &comparison) {
-            self.run.add_custom_comparison(comparison.as_str());
-
+        if self.run.add_custom_comparison(comparison.as_str()).is_ok() {
```
CryZe: Same here

src/run/editor/mod.rs
```diff
@@ -539,7 +536,7 @@ impl Editor {
             return Ok(());
         }
 
-        if validate_comparison_name(&self.run, new) {
+        if !new.starts_with("[Race]") && !self.run.comparisons().any(|c| c == new) {
```
CryZe: I'd like to have that logic in a single place. This is a copy of what's in add_custom_comparison. Maybe keep a validate_comparison_name somewhere and use it in both places.

src/run/parser/livesplit.rs
```diff
@@ -209,7 +209,7 @@ fn parse_segment<R: BufRead>(
                                 *segment.comparison_mut(&comparison) = t;
                             })?;
                         }
-                        run.add_custom_comparison(comparison);
+                        run.add_custom_comparison(comparison).ok();
```
CryZe: ? operator, probably needs a new error and .map_err(|_| Error::InvalidComparisonName).

src/run/parser/time_split_tracker.rs
```diff
@@ -102,7 +102,7 @@ pub fn parse<R: BufRead>(source: R, path_for_loading_other_files: Option<PathBuf
     parse_history(&mut run, path).ok();
 
     for comparison in comparisons {
-        run.add_custom_comparison(comparison);
+        run.add_custom_comparison(comparison).ok();
```
CryZe: same as above

src/run/parser/wsplit.rs
```diff
@@ -97,7 +97,7 @@ pub fn parse<R: BufRead>(source: R, load_icons: bool) -> Result<Run> {
     }
 
     if old_run_exists {
-        run.add_custom_comparison("Old Run");
+        run.add_custom_comparison("Old Run").ok();
```
CryZe: same as above

tests/run.rs
```diff
@@ -0,0 +1,48 @@
...
+
+mod run {
+    use livesplit_core::{Run, Segment};
+    use livesplit_core::run::Editor;
+
+    #[test]
+    fn add_comparison() {
+        let mut run = Run::new();
+        let c = run.add_custom_comparison("My Comparison").ok();
+        assert_eq!(c, Some(()));
```
CryZe: instead of using .ok() to convert to Option you can just assert against Ok(()), same for the other tests

src/run/parser/wsplit.rs
CryZe: actually possibly not, as "Old Run" should always be a valid comparison. So it should probably just be .expect(...)

src/run/parser/livesplit.rs
```diff
@@ -211,7 +211,8 @@ fn parse_segment<R: BufRead>(
                                 *segment.comparison_mut(&comparison) = t;
                             })?;
                         }
-                        run.add_custom_comparison(comparison);
+                        let _ = run.add_custom_comparison(comparison)
+                            .map_err(|_| Error::InvalidComparisonName);
```
CryZe: Is there a good reason why this just ignores the comparison?

src/run/mod.rs
```diff
@@ -33,3 +33,9 @@ pub use self::run::{ComparisonsIter, Run};
 pub use self::segment_history::SegmentHistory;
 pub use self::segment::Segment;
 pub use self::editor::Editor;
+
+/// Checks a given name against the current comparisons in the Run to
+/// ensure that it is valid for use.
+pub fn validate_comparison_name(run: &Run, new: &str) -> bool {
+    !new.starts_with("[Race]") && !run.comparisons().any(|c| c == new)
+}
```
CryZe: I'm not sure if it should be here. Maybe it should be in livesplit_core::comparison? Or maybe as a method on Run?

src/run/parser/time_split_tracker.rs
```diff
@@ -124,7 +128,8 @@ pub fn parse<R: BufRead>(source: R, path_for_loading_other_files: Option<PathBuf
     parse_history(&mut run, path).ok();
 
     for comparison in comparisons {
-        run.add_custom_comparison(comparison);
+        let _ = run.add_custom_comparison(comparison)
+            .map_err(|_| Error::InvalidComparisonName);
```
CryZe: The error is ignored here as well, but actually in cases like this (i.e third party format, which doesn't have these rules) it even should be ignored (or maybe even better: renamed). So then you don't need the InvalidComparisonName error.

CryZe:
The rest seems fine now :)

src/run/mod.rs
hallcristobal: I was thinking of putting I only put it in the `src/run/mod.rs` because the only two uses that I saw were children of Run (editor and Run itself) so I put it here. Would make sense to move it to comparison though.

src/run/parser/time_split_tracker.rs
hallcristobal: Now when you say "renamed" do you mean the new comparison name, so that it adds it regardless?

CryZe: Yeah, so that we can parse foreign formats that don't have these rules regardless.

CryZe: However you may need a proper Error type on add_custom_comparison then, that lets you distinguish a double add vs. a [Race] name

src/run/editor/mod.rs
```diff
@@ -662,26 +656,27 @@ impl Editor {
             return Ok(());
         }
 
-        if validate_comparison_name(&self.run, new) {
-            let position = self.run
-                .custom_comparisons()
-                .iter()
-                .position(|c| c == old)
-                .ok_or(())?;
+        match self.run.validate_comparison_name(new) {
```
CryZe: You don't do anything in the Err case, so you can just use ? instead of matching on it.

CryZe: Which also means that the return type of this function should have the proper Error type too.

CryZe: Also you have Result<T, ComparisonError> a lot. It may make sense to just provide a type def as ` ComparisonResult<T>`.

src/run/parser/livesplit.rs
```diff
@@ -211,7 +211,7 @@ fn parse_segment<R: BufRead>(
                                 *segment.comparison_mut(&comparison) = t;
                             })?;
                         }
-                        run.add_custom_comparison(comparison);
+                        run.add_custom_comparison(comparison).map_err(|_| {});
```
CryZe: Yeah, this definitely needs to be thrown. But I see why you can't: Because we add the same comparisons all the time. So you should ignore that certain error case for the livesplit parser (i.e. only throw the Race error). So that would be:
```rust
if let Err(ComparisonError::NameStartsWithRace) = run.add_comparison_name(comparison) { 
    return ... 
}
```

src/run/parser/time_split_tracker.rs
```diff
@@ -124,7 +125,21 @@ pub fn parse<R: BufRead>(source: R, path_for_loading_other_files: Option<PathBuf
     parse_history(&mut run, path).ok();
 
     for comparison in comparisons {
-        run.add_custom_comparison(comparison);
+        let mut name = comparison.to_string();
+        let mut number = 0;
+        loop {
+            match run.add_custom_comparison(name.clone()) {
```
CryZe: There's multiple problems with this approach. First, you can match on Err(ComparisonError::DuplicateName) instead of the ref e with if. Then you don't need the impossible case.

CryZe: Second, I noticed that we do this here way too late. The segments with the comparison times have already been added at this point. So there's a mismatch every time there's a rename like that. So you probably want to actually add these comparisons early (where we create the comparisons list up there) and do the inserting and renaming logic up there.

CryZe: Also the number renaming seems mostly fine to me. I'd probably start with 2 instead of 0 though ("Bla", "Bla2", "Bla3").

CryZe: I'm not sure if I like the `[ok]` though. I'd probably just cut away the [Race] and then try using that.

CryZe: Also I just noticed that in the duplicate name case, you actually keep adding the numbers, so it's ("Bla", "Bla2", "Bla23", "Bla234", ...), which is probably not what we want.

src/run/parser/xml_util.rs
```diff
@@ -35,6 +35,10 @@ quick_error! {
         ElementNotFound {}
         /// The length of a buffer was too large.
         LengthOutOfBounds {}
+        /// Parsed comparison has an invalid name.
+        InvalidComparisonName {
```
CryZe: InvalidComparisonName(err: ComparisonError)

src/run/run.rs
```diff
@@ -51,6 +51,16 @@ impl PartialEq for ComparisonGenerators {
     }
 }
 
+/// Error type for an invalid comparison name
+#[derive(PartialEq, Debug)]
+pub enum ComparisonError {
```
CryZe: You probably want to use quick_error! to define this. Look at the xml_util.rs code to see how to do that. That's also where you want to define the ComparisonResult type. (I think xml_util defines its own Result type too, so that's a good reference).

src/run/run.rs
```diff
@@ -341,10 +351,17 @@ impl Run {
...
     #[inline]
-    pub fn add_custom_comparison<S: Into<String>>(&mut self, comparison: S) {
+    pub fn add_custom_comparison<S: Into<String>>(
+        &mut self,
+        comparison: S,
+    ) -> Result<(), ComparisonError> {
         let comparison = comparison.into();
-        if !self.custom_comparisons.contains(&comparison) {
-            self.custom_comparisons.push(comparison);
+        match self.validate_comparison_name(&comparison) {
```
CryZe: You can use ? here instead of matching.

tests/run.rs
```diff
@@ -0,0 +1,48 @@
...
+    #[test]
+    fn rename_comparison_editor() {
+        let mut run = Run::new();
+        run.push_segment(Segment::new("s"));
+        run.add_custom_comparison("Custom").ok();
+        let mut editor = Editor::new(run).unwrap();
+        let c = editor.rename_comparison("Custom", "My Comparison");
+        assert_eq!(c, Ok(()));
+    }
+}
```
CryZe: I'm not sure but considering that there are bugs in the renaming logic atm, we may want to have unit tests for that? That sounds like the renaming logic should maybe even be refactored into a function in the comparison module then, so it can be tested as its own unit.

CryZe:
So the renaming should be like that "[Race] Bla" -> "Bla" -> "Bla2" -> "Bla3" -> ...  (this should probably be a unit test like that)

PR #89, opened 2017-12-18
Resume Run Functionality

Kigoli:
In order to accommodate the new Resume Run Functionality within LiveSplit, I needed to enhance LiveSplit-Core to look for and handle StopTime. StopTime is the time the timer was stopped at during a run.

### DIFF
--- a/capi/bind_gen/src/typescript.rs
+++ b/capi/bind_gen/src/typescript.rs
@@ -243,6 +243,7 @@ export interface RunEditorStateJson {
     icon_change: string | null,
     game: string,
     category: string,
+    stop_time: string,
     offset: string,
     attempts: number,
     timing_method: TimingMethodJson,
--- a/capi/src/run.rs
+++ b/capi/src/run.rs
@@ -180,6 +180,11 @@ pub unsafe extern "C" fn Run_metadata(this: *const Run) -> *const RunMetadata {
 }
 
 #[no_mangle]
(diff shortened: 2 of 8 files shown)


CryZe:
This breaks loading splits files that do not contain StopTime (so pretty much all of them). However more importantly, the entire actual implementation is missing. livesplit-core is an entire reimplementation of LiveSplit, so whatever is done in the original LiveSplit should be done here too. I don't blame you for missing this, as it's not entirely obvious. Your approach for resuming the run is an interesting approach, but I'm not quite sure if it's the one that makes the most sense, as it's intermingling attempts that are done and the actual active attempt. I didn't entirely verify your implementation yet, but I believe it also loses track of a few things, like what the game timer was at. So it may make more sense to introduce an entire `<ActiveAttempt>` section in the LSS, together with an ActiveAttempt type that tracks all this information properly. Also it may also make sense to store these active attempts in some temporary files instead of the LSS, as that is possibly only for actual historic information.

So overall I believe there's some larger architectural design to be had before this can be done properly. Additionally, it probably makes sense to figure out a design in livesplit-core, before getting it into the original LiveSplit.

CryZe:
Additionally this may actually also play into the eventual synchronization between different timers, so that should probably also be considered when figuring out a design.

CryZe:
I'm closing this in favor of a proper discussion happening over here: https://github.com/LiveSplit/livesplit-core/issues/156

PR #95, opened 2018-01-11
Implement Balanced PB

### LINKED ISSUE #82: Implement Balanced PB comparison and algorithm

CryZe:
Fixes #82 

## TODO
- [x] Implement the basic algorithm
- [x] Try to reuse / reduce as many allocations as possible (benchmark while doing so) (done, but still not too happy though)
- [x] Verify its correctness
- [x] Write a test that ensures an actual non-synthesized run works
- [x] Write doc comments
- [x] Possibly clean up and document some of its internals. The algorithm is really hard to understand.

### DIFF
--- a/src/comparison/balanced_pb.rs
+++ b/src/comparison/balanced_pb.rs
@@ -83,7 +83,7 @@ fn generate(
 
         // End early if we don't have any segment times anymore
         if weighted_segment_times.is_empty() {
-            len = i + 1;
+            len = i;
             break;
         }
 
--- a/src/comparison/balanced_pb.rs
+++ b/src/comparison/balanced_pb.rs
@@ -1,0 +1,214 @@
+use super::ComparisonGenerator;
+use {Attempt, Segment, TimeSpan, TimingMethod};
+use ordered_float::OrderedFloat;
+
+#[derive(Copy, Clone, Debug)]
+pub struct BalancedPB;
+
+/// The short name of this comparison. Suitable for situations where not a lot
+/// of space for text is available.
+pub const SHORT_NAME: &str = "Balanced";
+/// The name of this comparison.
+pub const NAME: &str = "Balanced PB";
+
+const WEIGHT: f64 = 0.9375;
+
+fn get_weight(index: usize, count: usize) -> f64 {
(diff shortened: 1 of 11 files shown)


src/comparison/balanced_pb.rs
```diff
@@ -0,0 +1,214 @@
...
+            let mut weighted_list = Vec::new();
+
+            if temp_list.len() > 1 {
+                temp_list
+                    .sort_unstable_by_key(|&(_, time)| OrderedFloat(time.total_milliseconds()));
+
+                let total_weight = temp_list.iter().map(|&(weight, _)| weight).sum::<f64>();
+                // TODO That's the smallest time's weight, not the actual
+                // smallest weight. Is that intended?
+                let smallest_weight = temp_list[0].0;
```
CryZe: @ROMaster2 We sort temp_list by the time, but then take the weight of the first element. That's the weight of the smallest time and not the smallest weight in general. Is that intended?

src/comparison/balanced_pb.rs
```diff
@@ -0,0 +1,214 @@
...
+                    agg_weight += weight;
+                    // TODO Possibly use extend instead for better reallocation behavior
+                    weighted_list.push((reweight(agg_weight, smallest_weight, range_weight), time));
+                }
+
+                // TODO What's the point of sorting this? temp_list was already
+                // sorted by the times and was inserted in that order.
+                // weighted_list was empty before.
+                weighted_list
+                    .sort_unstable_by_key(|&(_, time)| OrderedFloat(time.total_milliseconds()));
```
CryZe: @ROMaster2 this also seems very questionable. We sorted the temp_list before by the times and then store the elements in the weighted_list in the exact same order, so there doesn't seem to be a reason to sort them again.

CryZe:
Alright, I reused some allocations now.

Before:
```
test actual_splits ... bench:      92,692 ns/iter (+/- 6,330)
test fake_splits   ... bench:       5,709 ns/iter (+/- 359)
```

After:
```
test actual_splits ... bench:      80,635 ns/iter (+/- 5,624)
test fake_splits   ... bench:       4,517 ns/iter (+/- 466)
```

PR #103, opened 2018-02-06
Implement Parsing the original LiveSplit's layout files

CryZe:
This is still very much work in progress. And actually completely untested so far.

### DIFF
--- a/capi/bind_gen/src/emscripten.rs
+++ b/capi/bind_gen/src/emscripten.rs
@@ -619,6 +619,46 @@ const liveSplitCoreNative = {};"#
             }
         }
 
+        if class_name == "Layout" {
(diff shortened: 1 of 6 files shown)


TheEssenceOfDarkness:
When this is finished, how are we going to phase out the *.ls1l we use in LiveSplit One?

CryZe:
This is only for backwards compatibility and is only done on a best-effort basis, so this is going to be very lossy, as some things just don't apply 1:1 to LSO. So LSO will still always prefer its own layout format.

TheEssenceOfDarkness:
Oh I see. It is more of a thing for people switching to LiveSplit one, making it easier for people who decide to use it and wanting to port their old layouts over.

PR #106, opened 2018-02-14
Add SourceLiveTimer parsing

### LINKED ISSUE #105: Implement SourceLiveTimer Parsing
Seems to be valid to me, but ends up being accepted as a valid urn file which causes errors.  Zipped because github won't let you upload json files.

[test.json.zip](https://github.com/CryZe/livesplit-core/files/1722379/test.json.zip)

hallcristobal:
Resolves #105

### DIFF
--- a/src/run/parser/source_live_timer.rs
+++ b/src/run/parser/source_live_timer.rs
@@ -1,0 +1,104 @@
+//! Provides the parser for the SourceLiveTimer splits files.
+
+use std::io::Read;
+use std::result::Result as StdResult;
+use serde_json::de::from_reader;
+use serde_json::Error as JsonError;
+use {GameTime, Run, Segment, Time, TimeSpan};
+
+quick_error! {
+    /// The Error type for splits files that couldn't be parsed by the
+    /// SourceLiveTimer Parser.
+    #[derive(Debug)]
+    pub enum Error {
+        /// Failed to parse JSON.
+        Json(err: JsonError) {
+            from()
+        }
+    }
+}
+
+/// The Result type for the SourceLiveTimer parser.
+pub type Result<T> = StdResult<T, Error>;
+
+#[allow(non_snake_case)]
+#[derive(Deserialize)]
+struct Splits {
+    Category: String,
+    RunName: Option<String>,
+    Splits: Option<Vec<Split>>,
+}
+
+#[allow(non_snake_case)]
+#[derive(Deserialize)]
+struct Split {
+    Map: String,
+    Name: Option<String>,
+    Segment: Option<u64>,
+    BestSegment: Option<u64>,
+}
+
+fn time_span_from_ticks(category_name: &str, ticks: u64) -> TimeSpan {
+    let seconds = if category_name.starts_with("Portal 2") {
+        ticks as f64 / 30.0
+    } else {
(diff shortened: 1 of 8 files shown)


src/run/parser/source_live_timer.rs
```diff
@@ -0,0 +1,106 @@
+//! Provides the parser for the SourceLiveTimer splits files.
+
+#![allow(non_snake_case)]
```
CryZe: That allows the whole module to ignore snake case. It's probably better to have it on the two structs instead.

src/run/parser/source_live_timer.rs
```diff
@@ -0,0 +1,106 @@
...
+    RunName: Option<String>,
+    Splits: Option<Vec<Split>>,
+}
+
+#[derive(Deserialize)]
+struct Split {
+    Map: String,
+    Name: Option<String>,
+    #[allow(dead_code)]
+    Ticks: Option<u64>,
```
CryZe: If it's dead code, then I'd prefer it to be removed or even better, just commented out.

src/run/parser/source_live_timer.rs
```diff
@@ -0,0 +1,106 @@
...
+        run.set_game_name("Portal 2");
+    } else if splits.Category.starts_with("Portal") {
+        run.set_game_name("Portal");
+    } else if splits.Category.starts_with("Half Life 2") {
+        run.set_game_name("Half Life 2");
+    }
+
+    if let Some(run_name) = splits.RunName {
+        if run_name != splits.Category {
+            run.set_category_name(run_name);
```
CryZe: This seems to suggest that the run name is more likely to be the category name than the .Category field? Can you elaborate on that?

src/run/parser/source_live_timer.rs
hallcristobal: Makes sense, will change.

src/run/parser/source_live_timer.rs
hallcristobal: Only had this field here because of serde needing it for deserialization. Forgot there was an ignore tag

src/run/parser/source_live_timer.rs
hallcristobal: From looking at the source for SourceLiveTimer it appears that RunName can be changed, where as The CategoryName references a static name that's set by when making the splits, the first phrase of the String is the name of the game. If the RunName isn't changed by the user it defaults to the same as the CategoryName.

hallcristobal:
I will get a first try eventually.

src/run/parser/source_live_timer.rs
CryZe: doesn't it just ignore fields that you don't mention?

src/run/parser/source_live_timer.rs
CryZe: alright :)

src/run/parser/source_live_timer.rs
hallcristobal: Maybe? I remember when I tried (like 3 versions ago) it would throw an error, Expected ...

CryZe:
Looks like Travis is half dead atm. It seems like it compiles though, so I'll just merge it regardless.

PR #110, opened 2018-02-18
Correct SourceLiveTimer Parsing

### LINKED ISSUE #109: SourceLiveTimer files aren't correctly parsed
It looks like the split times are actually the segment times:

hallcristobal:
fixes #109

### DIFF
--- a/src/run/parser/source_live_timer.rs
+++ b/src/run/parser/source_live_timer.rs
@@ -34,7 +34,7 @@ struct Splits {
 struct Split {
     Map: String,
     Name: Option<String>,
-    Segment: Option<u64>,
+    Ticks: Option<u64>,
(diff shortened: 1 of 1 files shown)


CryZe:
Unfortunately GitHub won't let me inline comment there, but you have let best_split_time, which really should be best_segment_time and fall back to the split_time there, which doesn't make sense, cause one is segment time and one is a split time. So it should probably just be None instead of that fallback or something.

CryZe:
You don't use Time anymore, so you should remove that import.

CryZe:
I merged it manually https://github.com/LiveSplit/livesplit-core/commit/689a7e5d6042f00c9473109514f1cdf20653ce5c

PR #112, opened 2018-02-25
Python bindings: Automatically convert C strings to and from str type

AntyMew:
Currently, python function bindings taking or returning strings use a byte object `bytes` instead of string representation `str`. This is very unintuitive, leading to confusion when, for example, passing a string literal results in a `ctypes.ArgumentError`. Working around the issue forces relatively ugly code like this:
```python
run.set_game_name(b"Celeste")
run.set_category_name(b"Any%")
...
print(run.game_name.decode())
print(run.category_name.decode())
```
This fork cleans up API usage by running `str.encode` on every string argument and `bytes.decode` on every string return value.

### DIFF
--- a/capi/bind_gen/src/python.rs
+++ b/capi/bind_gen/src/python.rs
@@ -162,6 +162,8 @@ fn write_fn<W: Write>(mut writer: W, function: &Function) -> Result<()> {
                 "self.ptr".to_string()
             } else if typ.is_custom {
                 format!("{}.ptr", name)
+            } else if typ.name == "c_char" {
+                format!("{}.encode()", name)
             } else {
                 name.to_string()
             }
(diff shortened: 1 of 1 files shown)


CryZe:
Tagging @stevensmedia as I'm not entirely sure why this didn't happen before? Or is #69 and this changed from Python 2 to Python 3? I might check this later.

AntyMew:
It did happen before, actually. He used the workaround. See his repo [stevensmedia/pyvesplit-terminal](https://github.com/stevensmedia/pyvesplit-terminal/blob/master/pyvesplit-terminal). It has `encode('utf8')` and `decode('utf8')` littered throughout (functionally identical to `encode()` and `decode()`; 'utf-8' is the default encoding for both functions)

stevensmedia:
If I recall that's a python2 compat thing.  Been a while.

AntyMew:
To clarify... python2 `str` represents an ASCII string, and python3 `str` represents a UTF-8 string. Using `encode` and `decode` ensures that the bindings send and receive the same string format; `encode` always returns an ASCII string whether the input was ASCII or UTF-8, and similarly `decode` always returns a UTF-8 string.

Applying `encode` and `decode` to everything is a compromise between backwards-compatibility and performance. Both python2 and python3 will pass ASCII strings to the API, and receive UTF-8 strings back. I'm not sure if this creates any gotchas, but it ends with the same `print` behavior, at least. If we can get away with it we'd avoid a runtime check for the python version.

Code:
```python
from __future__ import print_function
from livesplit_core import Run

run = Run.new()
run.set_game_name("Celeste")
game = run.game_name()
print("{} ({})".format(game, type(game)))
```
Python2 output: `Celeste (<type 'unicode'>)`
Python3 output: `Celeste (<class 'str'>)`

PR #118, opened 2018-03-26
Correct Portal 2 Live Timer parsing

hallcristobal:
Correct Map Names & order
Handle merging of consecutive maps with the same name


CryZe:
Maybe we need something like cargo bomb / crater to parse all splits on splits i/o whenever we make changes to the parsers, so we know we don't break something.

CryZe:
Like for example you changed the order, but I'm pretty sure this used to be the correct order. But maybe not. Who knows.

hallcristobal:
It might have been at one point, i'm not sure I don't really follow Portal. I just know this is the order that I saw on https://wiki.sourceruns.org/wiki/Portal_2_Maps and in the PLT source

src/run/parser/portal2_live_timer.rs
```diff
@@ -159,18 +156,33 @@ pub fn parse<R: BufRead>(source: R) -> Result<Run> {
...
+                line = next_line;
+                let mut splits = line.split(',');
+                if splits.next() == Some(map) {
+                    let start_ticks: f64 = splits.next().ok_or(Error::ExpectedStartTicks)?.parse()?;
+                    let end_ticks: f64 = splits.next().ok_or(Error::ExpectedEndTicks)?.parse()?;
+                    let map_ticks = end_ticks - start_ticks;
+                    aggregate_ticks += map_ticks;
+                } else {
+                    break;
+                }
```
CryZe: I don't like the amount of duplication happening here. Can we turn this into a single loop maybe?

CryZe:
Also it looks like Splitter Bomb couldn't even find a single successfully uploaded Portal 2 Live Timer splits file on splits i/o. So I guess there's no proof it ever worked with any actual splits file.

PR #123, opened 2018-04-23
Implement Features 

### LINKED ISSUE #113: Introduce features for livesplit-core
We probably want these features:
 - parsing (one subfeature for each timer)
 - timing (includes Timer, SharedTimer, ...)
 - editing (includes RunEditor, ...)
 - components (includes the component module)
 - layouts (includes the layout module, forces the components feature)
 - analysis (includes the comparison and analysis module)

The C API and the binding generator are massively affected by this too.

hallcristobal:
WIP

Closes #113

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,10 +32,9 @@ members = ["capi", "capi/bind_gen", "capi/js", "capi/staticlib", "capi/cdylib"]
 base64 = "0.9.0"
 byteorder = "1.2.1"
 derive_more = "0.7.0"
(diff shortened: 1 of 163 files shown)


CryZe:
I think I'll close this. This is conflicting too much, and it's too big of a design change to easily commit to. I think introducing a few smaller and coarser features is probably the way to go. But maybe we also just want to split up livesplit-core into smaller crates? It's not entirely clear to me at this point. So this should probably all be discussed in the issue and then this can either be rebased and reopened in a new PR or entirely be redone.

PR #145, opened 2018-09-27
Shrink images that are too large

### LINKED ISSUE #47: Implement Image Resizing
It generally doesn't make much sense too store super large images, so we re-encode them as JPEGS < 256x256 (while keeping the aspect ratio in mind).

CryZe:
This is necessary eventually. However it also comes with a few downsides such as longer compile times, larger binary size and longer parsing times of splits due to the shrinking. This introduces the `image-shrinking` feature that allows you to choose whether you want to compile this in. It is activated by default. The plan is to introduce this to the original LiveSplit and Splits I/O before this lands in LiveSplit One, so that we can reduce the amount of actively used splits in the wild that still need to be shrunk once LiveSplit One gets the feature. 

Closes #47


Cargo.toml
```diff
@@ -49,3 +49,7 @@ chrono = { version = "0.4.0", features = ["serde", "clock"], default-features =
 
 [dev-dependencies]
 memmem = "0.1.1"
+
+[features]
+default = ["image-shrinking"]
+image-shrinking = ["image/webp", "image/pnm", "image/ico", "image/jpeg", "image/gif_codec", "image/tiff", "image/tga", "image/bmp", "image/hdr"]
```
glacials: Should `image/png` be here?

CryZe:
It's fairly slow opening splits that have large images in them, so I'd prefer reducing the amount of yet to be resized images that are around. I think natively compiled it does the conversion a bit faster, and we want to do it there as soon as possible anyway, so I'd rather wait a bit.

Cargo.toml
CryZe: image/png is always compiled in, as we already reencode Llanfair's headerless bitmaps to png. So these are only those that are conditionally activated.

PR #149, opened 2018-11-04
Implement Columns for the Splits Component

CryZe:
This implements initial support for columns for the splits component. The columns differ a bit from the original LiveSplit in that they don't have a single type. Instead they define a value they start out with and
a value they get updated with. This allows for some interesting combinations that were not possible before.


CryZe:
There's a few things that are still marked as TODO in there. So I'll either implement them before merging this or implement them in a followup PR.

PR #159, opened 2019-01-31
Improve handling the Best Segments comparison

CryZe:
The Best Segments comparison is often not suitable for deriving information as due to the shortest path calculation that we use for it, the difference between two segments may actually be meaningless. Instead we calculate information based on the stored best segment times for each individual segment and possibly add them up.


CryZe:
Superseded by #160

PR #163, opened 2019-02-10
Implement an abstract renderer

CryZe:
This PR adds a rendering module behind a new feature `rendering` to livesplit-core. The renderer here takes a LayoutState and an implementation of the Backend trait. The renderer takes the layout state and instructs the backend to creates meshes, textures and to render them out. By abstracting out the backend we can use this rendering on a lot of platforms without needing to focus on a specific rendering framework.

Additionally there's software rendering backend based on the euc crate behind the new `software-rendering` feature gate. This will allow us to unit test the rendering across a variety of platforms without needing to rely on a GPU framework.


CryZe:
This is still heavily WIP, as there's no unit tests yet, some documentation is missing and it could use some more refactoring.

PR #168, opened 2019-02-21
Implement Subsplits

### LINKED ISSUE #36: Implement Sub Splits into Splits Component
They should finally merge in livesplit-core.

CryZe:
This is a work in progress implementation of Subsplits. Instead of encoding the subsplits information into the segment names we extend the run format properly so the subsplits information can be stored. This is the roadmap for this Pull Request:
- [x] Parse the original modified segment names and rewrite them to be normal segment names.
- [x] Store the subsplits information in the Run structure
- [ ] Extend the splits component to handle the additional subsplits information.
  - [x] Implement basic support for subsplits.
  - [ ] Extend the "always show all splits" option to precalculate the maximum size necessary.
  - [ ] Add columns to the opened group headers.
  - [ ] Add setting for keeping previous group headers scrolled in.
  - [ ] Implement focus changing manual scrolling (scrolling should be able to open up other groups).
  - [ ] Handle columns that are based on relative times (such as segment times) on collapsed group headers. (e.g. it should be the combined segment time of the whole group, not just of the last segment in the group). Also add a test for this.
- [x] Extend the splits format to store the subsplits information.
- [x] Parse the subsplits information from the splits format.
  - [ ] Empty group names are the same as non-existing group names, this doesn't match the behavior elsewhere.
  - [ ] What to do about out of range groups?
  - [ ] There could be a group that is directly at the end such that it returns a view with an empty slice. That breaks a few assumptions.
- [ ] Extend the Run Editor to support the subsplits.
- [ ] Add Documentation.
- [ ] Add Tests.
  - [ ] Make sure the final split doesn't scroll the window back to the beginning.
- [ ] Improve horizontal support (such as the indentation not really making too much sense maybe)

Closes #36


capi/bind_gen/src/typescript.ts
```diff
@@ -244,11 +244,20 @@ export interface SplitStateJson {
      * on.
      */
     is_current_split: boolean,
+    /** Describes whether the segment is part of a segment group. */
+    is_subsplit: boolean,
+    /**
+     * Describes whether the row should be considered an even or an odd row.
+     * This is useful for visualizing the rows with alternating colors.
+     */
```
CryZe: This talks about rows, but in horizontal, we don't even use rows.

tests/run_files/segment_groups.lss
```diff
@@ -0,0 +1 @@
...up start="20" end="26"/><SegmentGroup start="26" end="33"/></SegmentGroups><AutoSplitterSettings><Splits><Split>Prologue</Split><Split>Chapter1Checkpoint1</Split><Split>Chapter1Checkpoint2</Split><Split>Chapter1</Split><Split>Chapter2Checkpoint1</Split><Split>Chapter2Checkpoint2</Split><Split>Chapter2</Split><Split>Chapter3Checkpoint1</Split><Split>Chapter3Checkpoint2</Split><Split>Chapter3Checkpoint3</Split><Split>Chapter3</Split><Split>Chapter4Checkpoint1</Split><Split>Chapter4Checkpoint2</Split><Split>Chapter4Checkpoint3</Split><Split>Chapter4</Split><Split>Chapter5Checkpoint1</Split><Split>Chapter5Checkpoint2</Split><Split>Chapter5Checkpoint3</Split><Split>Chapter5Checkpoint4</Split><Split>Chapter5</Split><Split>Chapter6Checkpoint1</Split><Split>Chapter6Checkpoint2</Split><Split>Chapter6Checkpoint3</Split><Split>Chapter6Checkpoint4</Split><Split>Chapter6Checkpoint5</Split><Split>Chapter6</Split><Split>Chapter7Checkpoint1</Split><Split>Chapter7Checkpoint2</Split><Split>Chapter7Checkpoint3</Split><Split>Chapter7Checkpoint4</Split><Split>Chapter7Checkpoint5</Split><Split>Chapter7Checkpoint6</Split><Split>Chapter7</Split><Split>Epilogue</Split></Splits></AutoSplitterSettings></Run>
```
CryZe: We might need to bump the version here (and everywhere else) if custom variables land first.

CryZe: They landed first. It's now 1.8.1.

PR #181, opened 2019-04-03
Add aarch64-pc-windows-msvc target

### LINKED ISSUE #165: Compile to aarch64-pc-windows-msvc
Support for that is landing soon in winapi: https://github.com/retep998/winapi-rs/issues/727

CryZe:
Closes #165

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,6 +9,8 @@ environment:
   matrix:
     # Aarch64
     - TARGET: aarch64-pc-windows-msvc
+      SKIP_TESTS: true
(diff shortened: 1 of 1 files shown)


CryZe:
This seems to now be blocked on backtrace not compiling (urgh why did everyone jump on failure, forcing backtrace as a dependency):
https://github.com/alexcrichton/backtrace-rs/issues/160

CryZe:
Apparently it tries to call link.exe but somehow the WSL /usr/bin/link (which is not even a linker, it's for symlinks) gets called instead.

PR #188, opened 2019-04-17
Implement Time Save Columns

### LINKED ISSUE #177: Add Possible Time Save to Column Start With
We may want to explore other possibilities there as well.

Requested here: https://twitter.com/Raging_Panic/status/1110837119243702272

CryZe:
There's now "Possible Time Save" as a Column Start With setting and "Time Saved or Lost" as a Column Update With setting.

- [x] Initial Implementation
- [x] Deduplicate "Segment Delta" and "Time Saved or Lost".
- [x] Tests

Resolves #177

### DIFF
--- a/capi/bind_gen/src/typescript.ts
+++ b/capi/bind_gen/src/typescript.ts
@@ -681,7 +681,11 @@ export interface CustomCombobox {
  * Specifies the value a segment starts out with before it gets replaced
  * with the current attempt's information when splitting.
  */
-export type ColumnStartWith = "Empty" | "ComparisonTime" | "ComparisonSegmentTime";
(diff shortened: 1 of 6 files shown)


CryZe:
This is kind of interesting. Apparently I completely forgot that the equivalent Update With value is "Segment Delta" and already exists. So at the moment "Time Saved or Lost" and "Segment Delta" are equivalent, except "Segment Delta" is a bad name, so I should remove "Time Saved or Lost" again and rename "Segment Delta".

wooferzfg:
I think it would be nice to have `column_possible_time_save_segment_delta` as an additional test (so that we're consistent with what we have for the other column types + we cover every valid use case)

PR #190, opened 2019-04-18
Set up CI with Azure Pipelines

CryZe:


CryZe:
Aaannnnd with that, there goes this PR.

PR #201, opened 2019-05-17
Implement Custom Variables

### LINKED ISSUE #200: Implement a Custom Variable System
We probably want to have a system that allows storing arbitrary key value pairs (probably both of type string initially). And then allow visualizing them as an info text component. The counter component can then use those to represent its inner state. The Run Editor can show them in the Metadata section. I'm guessing these variables should also have other information such as whether they are transient or are kept across attempts and whether they are read only in the Run Editor. Auto Splitters and components can provide them.

CryZe:
This Pull Request adds custom key value pairs to Runs. They can be created by the runner in the Run Editor and can store arbitrary string based values that the runner wishes to store with the splits. Components can then use these variables to visualize them. In particular, the text component can visualize these variables as key value pairs. Variables can also be provided by auto splitters, components or through other means. The automatic variables are temporary in nature and are therefore never stored in the splits files. The user also can not edit them. There also is an orthogonal scope per variable that binds the variable's value only to single attempts and resets its value between each one. This can be useful for per attempt counters.

Closes #200 

- [x] Rough prototype
- [x] Settings for text component
- [x] Saving custom variables
- [x] Loading custom variables
- [x] Adjustments based on proposal
- [x] Run Editor UI in LiveSplit One
- [x] Documentation
- [x] Tests


CryZe:
It seems like one of the remaining design decision is how we want to treat the "variable provider". So far the variable can be provided by the user in the run editor or it is automatically provided by some script, like an auto splitter. One problem here is that there is a conflict going on. What if the variable is user provided, but an auto splitter then tries to provide it as well. At the moment the automation is stronger and the variable is considered automated from that point on, meaning that it will not be stored in the splits file, definitely not optimal. Also we are lacking a notion of hybrid variables. Counters for example would be very much user provided and should be stored in the splits file (assuming the scope is "Always"). However some script is also automating their incrementation.

## Proposal

We call this the variable's `Lifetime`. A lifetime can be one of two values: `Temporary` or `Permanent`.
A permanent variable is always stored in the splits file and is never removed unless the user explicitly removes it via the Run Editor. A variable's lifetime can never be shortened - a permanent variable never becomes a temporary one, while a temporary variable can become a permanent one. If a variable is manually created in the Run Editor, it is always considered permanent. If the variable is created via a script, it is considered temporary by default, unless otherwise stated. The Run Editor only shows permanent variables. Creating a variable in the Run Editor that is already a temporary variable keeps its value.

CryZe:
@wooferzfg I'd like you to read through the proposal and want your opinion on it.

CryZe:
I almost feel like we should have design documents for every major design decision we do. And then have a test per sentence, that actually references the sentence of the design document.

wooferzfg:
- What's the functional difference between `Temporary` and `Permanent` variables? Is it just that `Temporary` variables don't show up in the Run Editor?
- Do we ever have a case where a `Permanent` variable is scoped `Per Attempt`? I can see where a `Permanent` variable would be scoped `Per Segment` (to save some sort of custom info), but `Per Attempt` seems inherently temporary to me. Also, "The value of an attempt scoped variable is not stored in the splits file and is not shown in the Run Editor" sounds like it's describing a `Temporary` variable.
- Even though we have an `Always` scope, I'm assuming that variables are saved per splits file. If so, we should consider the expected behavior when switching between splits files. What should components show if the variable it was displaying no longer exists?

CryZe:
> What's the functional difference between Temporary and Permanent variables? Is it just that Temporary variables don't show up in the Run Editor?
Temporary variables are usually provided by auto splitters and other scripts. So they are mostly statistical in nature and don't need to be saved. This also means you don't usually want to edit them.

> Do we ever have a case where a Permanent variable is scoped Per Attempt?
Mmh, I'd say that you create a death counter variable in the run editor, make it per attempt scoped and then it resets between attempts. However now that I'm thinking about it, we may not actually need to create it in the run editor. Just the layout settings or whatever might be sufficient.

> What should components show if the variable it was displaying no longer exists?
The text component for example simply shows a dash, just like other key value pair components (info text components) do. But yeah components that use variables need to be able to handle them not existing / having a value.

wooferzfg:
A death counter kinda seems like it should be a temporary variable.

Also, `Per Attempt` and `Per Segment` kinda seem like different types of scope to me. `Per Attempt` describes a one piece of information that gets cleared when out of scope, while `Per Segment` describes multiple pieces of information that are limited to each segment but don't actually get cleared (at least in the case of `Permanent` variables?).

CryZe:
No, they would get cleared the same way, but they could be stored in the segment history before being cleared. But every other variable could as well. So I don't think there's a fundamental difference there. Unless I'm misunderstanding what you are saying.

CryZe:
Maybe we should keep the notion of the scope out of this PR as this is really about the automation of when the variable should be cleared. But maybe that's something that individual scripts / components should decide themselves. Like setting the death counter to per attempt / per segment or whatever in the counter settings (not sure if it will be a component or whatever yet, but no matter what it will be, it will have settings)

wooferzfg:
Yeah, I can definitely work on this. We should talk about the UI design for these changes though.

CryZe:
@wooferzfg It probably makes sense to push to the branch in my PR instead of on the branch in this repo. (As you can see none of your changes show up as part of this PR).

Also in general when you are working on your own PRs, I'd rather prefer if you worked in a fork. The problem is when people fork the repo, they get a full snapshot of all the branches in it, which means they have a bunch of in-progress branches in their fork, that forever stay in-progress like that. I'd like to just have the master branch here (though dependabot sometimes creates a few branches here too, which I also don't like).

CryZe:
Not sure why CI is not running anymore?! Maybe it'll start after rebasing again.

PR #224, opened 2019-07-30
Port the PB Chance Component

### LINKED ISSUE #208: Port PB Chance Component
We might still want to do this. There already is an implementation here https://github.com/CryZe/livesplit-core/tree/pb-chance, but it was too numerically unstable to merge. Maybe we can take an approach similar to Balanced PB where we try to find the percentile that ends up with the PB. The percentile would then be the PB chance. One of the main problems of PB chance is that resets are problematic. I think Sethbling's version makes use of them. But if you do so, your PB chance is always very low, as most runs actually never make it to the end. However not considering the resets also is problematic as most runs that do not actually get reset are probably doing really well, so they are biased towards you actually getting a new Personal Best. However I believe we should go for a PB chance algorithm that ignores the resets as a low PB chance is discouraging. Encouraging the runner to reset is not something that LiveSplit should do. Also segments that haven't been reset aren't necessarily always overly good. We have the same bias in Average Segments, Median Segments and co. as well then.

CryZe:
This ports the PB Chance component from the original LiveSplit. Unlike the original component, this uses a different algorithm though. The livesplit-core version uses the same algorithm as the goal comparison generator and the Balanced PB comparison. The percentile chosen by the Balanced PB to reach the PB is the PB chance. Resets are ignored by this algorithm entirely.

Resolves #208

- [x] Basic algorithm
- [x] Tests
- [x] Documentation
- [x] Parse settings of the original PB chance component

I also sneaked in some other minor changes, such as more lifetime elision, removal of recursion limit and some more CI targets.


src/comparison/goal.rs
```diff
@@ -11,7 +11,8 @@ use ordered_float::OrderedFloat;
 /// The default name of the goal comparison.
 pub const NAME: &str = "Goal";
 
-const WEIGHT: f64 = 0.9375;
+const WEIGHT: f64 = 0.75;
```
CryZe: I changed the weight to match the other algorithms, such as median and average. I think we simply forgot to change this back when we lowered the value for the others. We might want to change this in the original LiveSplit too.

PR #236, opened 2019-09-06
Implement the splits.io API

### LINKED ISSUE #43: Implement the splits.io API

CryZe:
Closes #43


CryZe:
Rust 1.39 is stable now, so this can be merged now. It looks like we need to sufficiently guard the networking from being compiled in on certain platforms. This should've been obvious, but I never really thought about it.

CryZe:
This is blocked on the tokio ecosystem releasing non-preview crates now. This is tracked over here: https://github.com/LiveSplit/splits-io-api/issues/3

PR #245, opened 2019-09-13
Switch to GitHub Actions

CryZe:
This commit switches our CI from Travis CI + AppVeyor to GitHub Actions. There's a few minor problems still such as the Windows runners sometimes not being able to checkout the commit or upload the artifacts. This seems to be mitigated by running the Windows jobs first though. Other problems include that the currently released version of cross isn't compatible with GitHub Actions so we use a custom version in the meantime.


CryZe:
There's a couple of remaining issues with this:
 - Sometimes the network connection fails on the Windows jobs (checking out the commit or publishing a release).
 - You can't restart individual jobs, only the whole thing.
 - This uses a custom pre-release version of cross.
 - The 32-bit macOS build has been disabled. It doesn't seem to build. Supposedly 32-bit macOS is deprecated. Not sure if that's the problem though.
 - The asmjs build is disabled until the next cross version comes out.
 - FreeBSD is not supported anymore by cross.

PR #252, opened 2019-10-06
Reattach unattached Segment History Elements

### LINKED ISSUE #249: Implement a history fixing routine for unattached segment times
When you have segment time history elements that don't correspond to attempt history elements, through possibly manual modification of the split file, then the original LiveSplit just outright crashes in various ways, while livesplit-core seems to just overwrite the existing segment history elements when doing new attempts. Both seem suboptimal. There should probably be a fixing step that modifies the attempt index of the segment history elements with a new negative index.

This should be backported to the original LiveSplit.

https://splits.io/4of2

CryZe:
In `Run::fix_splits` we now check for segment history elements that are considered achieved by the runner, but don't correspond to an element in the attempt history. We move them over into the synthetic segment history elements (elements with index <= 0).

Fixes #249

### DIFF
--- a/src/run/mod.rs
+++ b/src/run/mod.rs
@@ -583,6 +583,7 @@ impl Run {
         }
         self.remove_duplicates();
         self.remove_none_values();
+        self.reattach_unattached_segment_history_elements();
     }
 
     /// Clears out the Attempt History and the Segment Histories of all the segments.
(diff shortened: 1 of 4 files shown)


CryZe:
The esmcripten failures are due to Rust having updated its emscripten version last night, so those are unrelated.

PR #257, opened 2019-10-14
Unify the Key Value based Component States

### LINKED ISSUE #255: Unify the States of the different Key Value Components
Key Value Components such as Previous Segment, Sum of Best Segments, Possible Time Save and so on all share the same visual appearance and could therefore use the same state objects. At the moment they are mostly the duplicates with very minor differences that could easily be unified. This simplifies the consumption of the component states such as in our internal abstract renderer, LiveSplit One and other frontends such as DarkSplit and glod.

This also contributes to #228 as it needs some amount of unification anyway. In fact most of these changes already happened in a prototype branch for that issue, so they can be extracted out to solve this issue.

CryZe:
They used to mostly be copies of each other's code. Since they mostly share visual appearance, this deduplicates that code such that they share a single key value component state that encompasses all the various ways they can be visualized. This should also simplify adding even more key value based components - long term even allowing for custom components provided by user scripts.

Resolves #255


CryZe:
The emscripten fails are failing due to ryu: https://github.com/dtolnay/ryu/issues/18
Windows gnu Nightly is failing due to: https://github.com/rust-lang/rust/issues/65335
And clippy is not reporting its stuff due to: https://github.com/actions-rs/clippy-check/issues/2

Oof

PR #264, opened 2019-10-30
Implement a segment time component.

### LINKED ISSUE #259: Possibly implement a Segment Time Component
Unless I'm missing something, I don't think you can currently show the Segment Time of the current comparison / the comparison you selected. We have Previous Segment, which is the Segment Delta and the Delta Component, but no Segment Time component. Someone on r/speedrun suggested adding more segment times to the Detailed Timer, but that would probably get too cluttered, so having Segment Time as its own component would probably make sense.

TheEssenceOfDarkness:
This draft PR contains initial work for implementing a segment time component. This is far from merge ready right now but the final plan is to have it act similarly to the segment time display on the detailed timer. 

One thing that needs to be decided on is the name of the key for display on the layout. Right now it is simply named "Segment Time" but I feel like this could be a lot more descriptive. Currently my plan is to show the comparison in parenthesis if it differs from the timer's current comparison.

Also I'm not sure if I documented everything correctly either.

Closes #259


.gitignore
```diff
@@ -1,3 +1,4 @@
+**.patch
```
TheEssenceOfDarkness: This is only here to prevent accidentally committing patches in case you were wondering.

CryZe: There's a .git/info/exclude file that acts as a local .gitignore in every git repository.

TheEssenceOfDarkness: i'll remove this in the next commit and add it there.
thanks for letting me know

TheEssenceOfDarkness:
>One thing that needs to be decided on is the name of the key for display on the layout. Right now it is simply named "Segment Time" but I feel like this could be a lot more descriptive. Currently my plan is to show the comparison in parenthesis if it differs from the timer's current comparison.

For this I've decided to just do what the Previous Segment component does here.

TheEssenceOfDarkness:
The component itself is fully functional now but some code cleanup is needed.

The last thing we should ask is if we should provide a timing method override.
I'm split between providing one to match the detailed timer or not providing one for consistency with the other components.

CryZe:
I believe we want to have more specialized key names, such as "Best Segment Time", "Worst Segment Time", "Average Segment Time" (and possibly more).

CryZe:
I added a commit with some of these changes and TODO comments about changes that still need to be done.

src/component/segment_time.rs
```diff
@@ -0,0 +1,214 @@
...
+                .segment(last_split_index)
+                .comparison_timing_method(comparison, timing_method)?
+                - run
+                    .segment(last_split_index - 1)
+                    .comparison_timing_method(comparison, timing_method)?,
+        )
+    } else {
+        None
+    }
+}
```
TheEssenceOfDarkness: There's some stuff I'm still unsure about here:
- If this should go under `crate::analysis::state_helper`
- The fix for split indexes larger than `run.len() - 1` causing panics. (lines 197-201)
- Code formatting.

src/component/segment_time.rs
```diff
@@ -0,0 +1,214 @@
...
+
+impl Default for Settings {
+    fn default() -> Self {
+        Self {
+            background: key_value::DEFAULT_GRADIENT,
+            comparison_override: None,
+            display_two_rows: false,
+            label_color: None,
+            value_color: None,
+            accuracy: Accuracy::Hundredths, // TODO: Get this default from the formatter (some new constant).
```
TheEssenceOfDarkness: I'm also very unsure how you'd want me to go about doing this.
All of the other components have the accuracy specified directly rather than getting it from the formatter.

wooferzfg:
I think it might be good to have an option (that is on by default) to switch to a live segment time whenever your current segment is longer than the comparison segment. That would make this component consistent with pretty much every other component we have.

TheEssenceOfDarkness:
>I think it might be good to have an option (that is on by default) to switch to a live segment time whenever your current segment is longer than the comparison segment.

The original intent here was to be identical to how the segment times are on the detailed timer but that might not be a bad thing to add.

src/component/segment_time.rs
TheEssenceOfDarkness: > If this should go under `crate::analysis::state_helper`

That does in fact seem to be the best place for it, but the name you suggested conflicts with another function in the detailed timer.

> If the fix for split indexes larger than `run.len() - 1` causing panics is acceptable. (lines 197-201)

It wasn't even working properly, but I fixed that in my latest commit.

> General code formatting.

Honestly I don't think there is much I can do in that regard.

TheEssenceOfDarkness:
The only thing I'm stuck on now is moving `calculate_comparison_time` into the analysis module because I suck at naming stuff. Otherwise it would basically be done aside from the feature @wooferzfg suggested.

The suggested name of `calculate_segment_time` sounds too similar to stuff already defined in `state_helper.rs` while also conflicting with a function defined in the detailed timer.

TheEssenceOfDarkness:
@CryZe, I hit a bit of a roadblock when it comes to moving `calculate_comparison_time` into the analysis module and I want your opinion on the best course of action to take here.

Right now I've have 3 good options:
1. Remove `calculate_comparison_time` from the component entirely and use `comparison_segment_time` from the analysis module.
2. Move it into the analysis module under the name `calculate_segment_time`
3. Leave it as-is.

Option 1 seems like the best course of action but it won't do exactly what we want. (It'll combine segments if there is an empty split instead of returning None)

Option 2 results in 2 functions with almost identical purpose in the analysis module. In addition, the name also conflicts with something defined in the detailed timer so we'll have to rename that.

Option 3 just results in duplicated code.

src/analysis/state_helper.rs
```diff
@@ -69,6 +69,12 @@ pub fn comparison_segment_time(
     comparison: &str,
     method: TimingMethod,
 ) -> Option<TimeSpan> {
+    let segment_index = if segment_index >= run.len() {
+        run.len() - 1
+    } else {
+        segment_index
+    };
+
```
CryZe: This definitely shouldn't be here. The detailed timer only does this because we still show the segment you just finished at the end of the run. No other component does this though, so the segment time component shouldn't do this either. (We may want to change this, but that should be done for all components in a separate pull request then).

CryZe: The detailed timer does additional stuff in its equivalent function that was used before, such as choosing the best segment time instead of subtracting the split times for the best segments comparison. That should probably be done here too, such that the detailed timer can use this instead of its own copy. (cc @wooferzfg)

CryZe: That obviously changes the behavior in the splits component, but that's probably what we want anyway. That way it's consistent across the segment time component, detailer timer and the splits component.

wooferzfg: Yeah, it would be a lot better if these calculations were standardized across components.

However, I'm not sure I agree that this component shouldn't show the last segment's time after the run is over. The previous segment and delta components both show info about the last segment once the run ends. Possible time save doesn't show any info, but that's arguably not the behavior that we want. I would prefer to always show info about the last segment when the run ends rather than just showing nothing.

CryZe: I think we should still postpone that until it‘s clear which components could benefit from it. I‘d rather land this without the logic for now and we can decide what to do in subsequent PRs.

TheEssenceOfDarkness: @CryZe, I added this logic here because it prevents an index out of range panic if the caller were to pass a split index `>= run.len()`. This has no effect on the behavior of the splits component as its just clamping the split index to `run.len() - 1`.

Either way this logic needs to be *somewhere*, either here, or in the component itself otherwise LSC will panic after the run ends.

CryZe: It should be in the component. It shouldn‘t query the information for the current segment‘s time if there‘s no current segment. Handling wrong inputs by just clamping them is generally just causing hidden bugs.

TheEssenceOfDarkness: I moved the logic into the component and documented the panic.

src/analysis/state_helper.rs
```diff
@@ -63,6 +63,11 @@ fn find_previous_non_empty_split_and_comparison_time(
...
 /// method specified. This is not calculating the current attempt's segment
 /// times.
+///
+/// # Panics
+///
+/// Panics if the provided `segment_index` is greater than or equal to
+/// `run.len()`.
 pub fn comparison_segment_time(
     run: &Run,
     segment_index: usize,
```
CryZe: This function should have the additional checks that the detailed timer is doing, such as the specialized logic for the best segments comparison and so on. The goal should be to remove the function from the detailed timer and use this one instead. Then it should be good to merge.

CryZe: @wooferzfg Looks like in the splits we use "combined comparison segment time", in the detailed timer we use "single comparison segment time" and for the segment time component it could potentially be either. What do you think we should do? Do the same thing for all of them?

wooferzfg: My initial reasoning for using the combined segment time in the splits vs. a single segment time in the detailed timer is that you can see all of the segments in the splits, which means you know which segment the combined segment time started at. In the detailed timer, you can only see the current segment, so we never show combined segments. There's also the issue of the segment timer starting at a different segment than the comparison segment time if we start combining stuff.

However, we don't really apply this logic everywhere else, so it's not totally clear what we should do here. The previous segment component uses combined segments, for example. In my opinion, it would be best to use the combined comparison segment time here because there's only one piece of information being displayed.

CryZe: Yeah I think for the splits it makes sense to keep it showing the combined segments, as you can see that the ones before are skipped. For the detailed timer you don't, so it makes sense to dash it out. But I'm not quite following your argument for the segment time component, as you seem to argue the opposite there?
> combined comparison segment time here because there's only one piece of information being displayed.

wooferzfg: For the segment time component, we only show one value, so there's no issue with conflicting information within the component (unlike the detailed timer, where we show a segment timer and 2 different comparisons). I don't feel too strongly about this though.

TheEssenceOfDarkness: I don't have many strong feelings here either, but if I was forced to choose I'd say just unify them for the sake of code quality and consistency.

TheEssenceOfDarkness:
Looks like this is should be good to merge now. Before we merge this, I'm going to squash 241ebf0 and 81faae3 so the commit history is cleaner.

src/analysis/state_helper.rs
```diff
@@ -63,12 +63,21 @@ fn find_previous_non_empty_split_and_comparison_time(
 /// Calculates the comparison's segment time of the segment with the timing
 /// method specified. This is not calculating the current attempt's segment
 /// times.
+///
+/// # Panics
+///
+/// Panics if the provided `segment_index` is greater than or equal to
+/// `run.len()`.
 pub fn comparison_segment_time(
```
wooferzfg: I don't think that the current implementation of this method makes sense.

- This provides a combined segment time, which means that the detailed timer's logic will change when switching over to it.
- Using the best segment time only makes sense if we're using non-combined segment times (because, by definition, the best segment time is a non-combined segment time). That's why the detailed timer was using it previously, but the splits component was not.

CryZe: Yeah, this wasn't meant to be finished.

wooferzfg: I would prefer if there were two separate methods in `state_helper`: one for combined segment times and one for non-combined segment times. The non-combined segment time method would do the best segment time override, while the combined segment time method would not. The splits component would use the combined segment time method, and the detailed timer component would use the non-combined segment time method. The segment time component would use whichever one we think makes the most sense (which is the discussion we just had in the other thread that didn't reach a clear answer).

CryZe: So if we end up splitting comparison segment time into two distinct methods and splits keep using the combined one, then there's no way for the splits to show the best segment times as a column. Are we sure that's what we want? Is there a disadvantage to just switch to the best segment times despite us usually showing combined segment times? The problem is that just like best split times and co. we can't really derive any segment times from it, as the times aren't truly building on top of each other.

src/timing/formatter/segment_time.rs
```diff
@@ -7,12 +7,12 @@ pub struct Inner {
...
 
-/// The Possible Time Save Time Formatter formats Time Spans for them to be
-/// shown as Time Saves. This specifically means that the fractional part of the
-/// time is always shown and the minutes and hours are only shown when
-/// necessary. The default accuracy is to show 2 digits of the fractional part,
-/// but this can be configured. Unlike the Short Time Formatter, the Possible
-/// Time Save Formatter shows a dash when there's an empty time.
+/// The Segment Time Formatter formats Time Spans for them to be shown as
+/// Segment Times. This specifically means that the fractional part of the time
+/// is always shown and the minutes and hours are only shown when necessary. The
```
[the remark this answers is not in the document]
TheEssenceOfDarkness: Honestly we should just remove it entirely. Its equivalent to using the short time formatter with the `DashWrapper` formatter.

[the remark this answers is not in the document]
CryZe: No, then we are getting inconsistent time formatting across the components. We can probably clean this up a bit long term, but I'd also like to keep this out of this PR. Segment Time is close enough to what this is used for, at least most of the time. If someone comes up with a better name, we can do it as part of this PR, but any larger changes shouldn't be part of this PR.

src/analysis/state_helper.rs
TheEssenceOfDarkness: Personally I'd prefer just having a single method and explicitly handling empty segments in the detailed timer itself if the resulting code is cleaner than having 2 different methods with almost identical code.

wooferzfg: Eh, I guess it's fine to use the best segment times then, even if it's a little inconsistent.

TheEssenceOfDarkness: Looks like we might actually have to have 2 different functions.

The cleanest way I can think of handling things in the detailed timer is to use a wrapper function and at that point we might as well just have 2 functions.

TheEssenceOfDarkness: In the end I went with the solution of having 2 different methods. I couldn't find a way to cleanly keep the current behavior of the detailed timer besides creating a wrapper function.

CryZe:
I'll try to review this on the weekend.

TheEssenceOfDarkness:
@CryZe, This has been sitting idle for about 2 weeks so I'm wondering if you have found time to review this yet.

Issue #275, opened 2019-11-30
Make it feasible to use the component states via the C API without serializing


TheEssenceOfDarkness:
Using `#[repr(C)]` structs for certain data types could really improve the usability of the C API compared to parsing JSON.

I have very little ideas for what this could look like so I'm going to experiment a bit here. We will definitely need to add support for this in the bindings generator.

This should be labeled as low priority as this isn't *too* important but would definitely improve API usability. 

General Idea:

```rust
#[repr(C)]
struct Data {
    x: u32,
    y: u32,
}
```
would become
```c
struct Data {
    uint32_t x;
    uint32_t y;
};
```

The libc crate will probably be extremely useful here.

CryZe:
I don't think that's feasible. It's important that we can generate bindings for all the languages that we are currently generating bindings for. So this would mean that we need to figure out how to express C structs in all of those languages (which contain scripting languages like Python, Ruby, JavaScript and co. which will complicate this) and also we need to annotate everything with repr(C), which we can't even do for everything. Any HashMaps, Vecs and co. will make this basically impossible.

So what's much more likely is that we are staying OOP, not leak any implementation details and have methods for accessing all the data like we already do in the C API. This just needs to be done. However we will then run into another problem (indirectly) and that is that the C API is getting extremely big. To the point where it significantly hurts the web version of LSO. So we then likely also needs to figure out how to conditionally compile in subsets of the API, instead of the whole thing. But that needs to be understood by the binding generator then (it's not that hard to do this and we need it anyway for networking and co.).

TheEssenceOfDarkness:
Personally I don't think making everything in livesplit-core `#[repr(C)]` is the right approach here to begin with, and the C API having hundreds of getter functions isn't good either. A good solution would involve having the C API define its own boxed C structs and passing them around. We could then return a lot more data from a single function call then we'd be able to otherwise.

I experimented a bit with this outside of livesplit-core and [it seems to work a lot better than I had imagined.](https://gist.github.com/TheEssenceOfDarkness/f9c9d1b60c84b081f8169e1282cd3890)


---

I've also started looking for the documentation we'll need to support this in ever language we currently generate bindings for.

- [x] C/C++ - Native Support
- [ ] C# - Supported but can't find documentation for it yet.
- [x] Java (JNA) - [Documented Here](http://java-native-access.github.io/jna/5.4.0/javadoc/overview-summary.html#structures)
- [ ] Java (JNI)
[...]

PR #278, opened 2019-12-16
C API: Support generating definitions for structs.

TheEssenceOfDarkness:
# Overview

The goal of this PR is to support generating definitions for `#[repr(C)]` structs defined in the C API. The main use case behind this will be returning a component's state as a single struct instead of going through JSON or multiple getter functions.

This will definitely require some experimentation and discussion so I don't see this getting merged any time soon. Most of the discussion for this is happening in #275

# Todo
## Needed Improvements
- [ ] Only generate definitions for structs marked `#[repr(C)]`
- [x] Not generating definitions in languages where we don't support this. (Obsoleted by making this an extension to the opaque API)
- [x] Support documentation.
- [ ] Find a way to attach a struct to a class instead of having it as a separate datatype. 

## Languages to Support:
- [x] C
- [ ] C# (Requires unsafe code)
- [ ] Java/Kotlin JNI (Unlikely to happen)
- [ ] Java JNA
- [ ] JS/TS (Emscripten)
- [ ] JS/TS (Node FFI)
- [ ] JS/TS (Wasm)
- [ ] Ruby
- [ ] Python

### DIFF
--- a/capi/bind_gen/src/c.rs
+++ b/capi/bind_gen/src/c.rs
@@ -1,4 +1,4 @@
-use crate::{Class, Type, TypeKind};
+use crate::{Class, Struct, Type, TypeKind};
 use std::borrow::Cow;
 use std::collections::BTreeMap;
 use std::io::{Result, Write};
@@ -35,7 +35,11 @@ fn get_type(ty: &Type) -> Cow<'_, str> {
     name
 }
 
-pub fn write<W: Write>(mut writer: W, classes: &BTreeMap<String, Class>) -> Result<()> {
+pub fn write<W: Write>(
+    mut writer: W,
+    classes: &BTreeMap<String, Class>,
+    structs: &Vec<Struct>,
+) -> Result<()> {
     write!(
         writer,
         "{}",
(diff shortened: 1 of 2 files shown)


TheEssenceOfDarkness:
Before I continue work on this, I'm going to spend some time working out how we should actually make use of this if and when it's finished.

I'm kind of stalled on how we should represent generic data types and variable length arrays. (e.g. vectors).

TheEssenceOfDarkness:
I'm going to abandon this in favor just of having getter functions for a couple of reasons:
- This will be *really* difficult to support in wasm.
- This will make the bindings generator even more hacky than it has to be
- Because there are languages where we can't support it, we'll have extra API bloat from having to have fallback getter functions just to access the structs.
- It is now very reasonable to just serialize a component state to a struct manually due the fact that we are now able to directly access them from a layout state anyway. (#289)

```c
//example of manual serialization (probably incorrect but the idea is there)
struct KeyValueState {
    char *key;
    char *value;
};

struct KeyValueState *serialize(KeyValueComponentStateRef self)
{
   struct KeyValueState *ret = malloc(sizeof(struct KeyValueState));
   ret->key = strdup(KeyValueComponentState_key(self));
   ret->value = strdup(KeyValueComponentState_value(self));
   return ret;
}

void drop(struct KeyValueState *ptr) {
    free(ptr->key);
    free(ptr->value);
    free(ptr);
}
```

PR #289, opened 2020-01-20
capi: Expose Layout State object

### LINKED ISSUE #275: Make it feasible to use the component states via the C API without serializing

TheEssenceOfDarkness:
This exposes the Layout State object to the C API making it somewhat feasible to get the layout state without going though JSON. Unfortunately there is not a good type safe way to do this, so it is up to the person using the library to call the correct methods for accessing the component states.

Generally the intended usage should look about like this:
```
- Query the length of the LayoutState
- Do the following for every component in the layout's state:
    - Get the type of the component.
    - Use said type information to call the correct method to get
      the components state. 
```

There is also some stuff that is up for debate here:
- Method naming
- If we actually need to expose the Separator component's "state".

Merging will close #275 as this is sufficient for reading component states without serializing, and missing properties can easily be added.


capi/src/layout_state.rs
```diff
@@ -0,0 +1,159 @@
...
+
+use livesplit_core::component::blank_space::State as BlankSpaceComponentState;
+use livesplit_core::component::detailed_timer::State as DetailedTimerComponentState;
+use livesplit_core::component::graph::State as GraphComponentState;
+use livesplit_core::component::key_value::State as KeyValueComponentState;
+use livesplit_core::component::separator::State as SeparatorComponentState;
+use livesplit_core::component::splits::State as SplitsComponentState;
+use livesplit_core::component::text::State as TextComponentState;
+use livesplit_core::component::timer::State as TimerComponentState;
+use livesplit_core::component::title::State as TitleComponentState;
```
CryZe: you can do `use livesplit_core::component::{ ... };` here

capi/src/layout_state.rs
```diff
@@ -0,0 +1,159 @@
...
+        ComponentState::BlankSpace(_) => b"BlankSpace\0".as_ptr(),
+        ComponentState::DetailedTimer(_) => b"DetailedTimer\0".as_ptr(),
+        ComponentState::Graph(_) => b"Graph\0".as_ptr(),
+        ComponentState::KeyValue(_) => b"KeyValue\0".as_ptr(),
+        ComponentState::Separator(_) => b"Separator\0".as_ptr(),
+        ComponentState::Splits(_) => b"Splits\0".as_ptr(),
+        ComponentState::Text(_) => b"Text\0".as_ptr(),
+        ComponentState::Timer(_) => b"Timer\0".as_ptr(),
+        ComponentState::Title(_) => b"Title\0".as_ptr(),
+    }) as *const c_char
```
CryZe: I think you can move the `.as_ptr()` here, and even do `.as_ptr().cast()` then.

capi/src/layout_state.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+/// Gets the Blank Space component state at the specified index.
+#[no_mangle]
+pub extern "C" fn LayoutState_component_as_blank_space(
+    this: &LayoutState,
+    index: usize,
+) -> &BlankSpaceComponentState {
+    match &this.components[index] {
+        ComponentState::BlankSpace(x) => &x,
```
CryZe: Since you matched on a reference, the x already is a reference, no need to do &x again.

capi/src/layout_state.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+/// Gets the Detailed Timer component state at the specified index.
+#[no_mangle]
+pub extern "C" fn LayoutState_component_as_detailed_timer(
+    this: &LayoutState,
+    index: usize,
+) -> &DetailedTimerComponentState {
+    match &this.components[index] {
+        ComponentState::DetailedTimer(x) => &x,
```
CryZe: Same here and everywhere below.

capi/src/layout_state.rs
TheEssenceOfDarkness: I've tried moving the `.as_ptr()` calls out of the match statement. Unfortunately, rustc complains that the strings are not the same type due to them being different lengths.

However, using `.cast()` does work so I'm going to use it here.

CryZe: Ah, I guess you could use normal str literals (so without the b-prefix) and then it should work.

TheEssenceOfDarkness:
I've addressed all of things you've wanted me to change.

CryZe:
Can you try the string literals instead of the raw byte literals?

TheEssenceOfDarkness:
You replied to that really quickly. The string literals did work.

CryZe:
Looks like it's not properly formatted, but I'll merge it anyway.

PR #292, opened 2020-01-21
Use img_hash for rendering tests

### LINKED ISSUE #287: Use perceptual hashing algorithm for the rendering tests
Looks like we are getting some floating point inconsistencies on x86_64-unknown-linux-musl now. We already had issues on i586 before where we disabled the tests, but we should probably do a more robust hashing algorithm suited for images where tiny differences don't cause the hash to change.

wooferzfg:
Fix #287

### DIFF
--- a/src/rendering/software/mod.rs
+++ b/src/rendering/software/mod.rs
@@ -2,30 +2,7 @@
 //! is much slower than with a normal GPU, but might be sufficient for
 //! situations where you want to create a screenshot of the layout.
 
-#[cfg(all(
-    test,
-    not(any(
-        // Without MMX the floating point calculations don't follow IEEE 754. So the
-        // tests fail to produce the correct checksums.
-        // FIXME: We use SSE as an approximation for the cfg, because MMX isn't
-        // supported by Rust just yet.
-        all(
-            target_arch = "x86",
-            not(target_feature = "sse"),
-        ),
-        // Additionally the x86_64-unknown-linux-musl target seems to be failing
-        // with lyon 0.15. I'd assume it's because of musl's libm doing
-        // something different than the glibc one, but I'm honestly not sure.
-        // FIXME: We likely want a perceptual image hash instead, which would be
-        // much more robust.
-        // https://github.com/LiveSplit/livesplit-core/issues/287
-        all(
-            target_os = "linux",
-            target_arch = "x86_64",
-            target_env = "musl",
-        ),
-    )),
-))]
+#[cfg(test)]
 mod tests;
 
 use {
--- a/src/rendering/software/tests.rs
+++ b/src/rendering/software/tests.rs
@@ -6,8 +6,8 @@ use {
         run::parser::{livesplit, llanfair, wsplit},
         tests_helper, Run, Timer,
     },
-    crc::crc32::checksum_ieee,
     image::Rgba,
+    img_hash::{HasherConfig, ImageHash},
     std::{
         fs::{self, File},
         io::BufReader,
@@ -40,7 +40,7 @@ fn default() {
 
     let state = layout.state(&timer);
 
-    check(&state, 0x9df41392, "default");
+    check(&state, "luCCVRJIPLE=", "default");
 }
 
 #[test]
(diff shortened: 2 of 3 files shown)


wooferzfg:
@CryZe The `single_line_title` rendering test seems to produce different results in the CI than it does for me locally. This issue also seems to be happening in other PRs, where I think the test randomly started failing over the last few days.

src/rendering/software/mod.rs
```diff
@@ -4,26 +4,14 @@
...
-            target_env = "musl",
-        ),
+
+    // Without MMX the floating point calculations don't follow IEEE 754. So the
+    // tests fail to produce the correct checksums.
+    // FIXME: We use SSE as an approximation for the cfg, because MMX isn't
+    // supported by Rust just yet.
+    not(all(
+        target_arch = "x86",
+        not(target_feature = "sse"),
```
CryZe: If the image hash is good enough, we shouldn't need this anymore either.

CryZe:
> The single_line_title rendering test seems to produce different results in the CI than it does for me locally.

Even now with the image hash? You may need to run `cargo update` then to get the latest versions of each crate. But yeah I noticed that this changed recently as well.

wooferzfg:
The tests pass for me after running `cargo update`. The difference between the hash I was getting before and the one I get now is just 1 bit, so we could potentially have a tolerance for the distance between hashes that's more than 0. I'm just not sure what that tolerance should actually be.

CryZe:
I think keeping it on 0 for now is okay. Because the image we are getting now is actually fairly questionable. We are seeing some weird diagonal aliased line now, which may be worth investigating and reporting to whatever caused this. So I feel like we shouldn't just completely brush off notable changes like these via a larger tolerance value.

src/rendering/software/mod.rs
wooferzfg: Seems like some of the image comparisons are still slightly off. Could we retrieve the images from the targets that are failing and see what it actually looks like?

CryZe:
Well, looks like we need a tolerance value on non-SSE platforms at least. Maybe we can use the cfg we had previously to conditionally give those platforms a larger tolerance value.

src/rendering/software/mod.rs
CryZe: You could try to run these tests locally. Just install the specific target and run the tests with the target.

CryZe:
Another way to minimize these problems should be to render the more broken tests in a higher resolution. That way minimal changes such as one glyph being +/- 1 pixel off causes less of a relative error.

wooferzfg:
I switched to using `actual` and `expected` for the file names because:
- I was having trouble differentiating which file was which when the hashes were too similar, which will happen much more often with perceptual hashing.
- base64 uses `/`, which I think causes problems when creating the file path.

wooferzfg:
I tried using bigger hashes, but that caused the rendering tests on the Linux mips builds to start failing for some reason.

src/rendering/software/tests.rs
```diff
@@ -200,26 +200,46 @@ fn horizontal() {
...
+    let distance = calculated_hash.dist(&expected_hash);
 
     fs::create_dir_all("target/renders").ok();
 
-    let path = format!("target/renders/{}_{:08x}.png", name, calculated_checksum);
-    image.save(&path).ok();
+    let expected_path = format!("target/renders/{}_expected.png", name);
+
+    if distance > get_comparison_tolerance() {
+        let actual_path = format!("target/renders/{}_actual.png", name);
```
CryZe: I'm not sure if I'm too much of a fan of this. The old one had the advantage that I could switch branches and it still would find the expected image for the specific branch, so if visual changes would happen, it's able to create a proper diff.

CryZe: What we could do is throw the hash bytes into AES to get highly scrambled bytes out again and revert the AES to get back the image hash for the expected AES hash.

TheEssenceOfDarkness: what stops us from using hex instead of base64?

wooferzfg: For now, I reverted to the old behavior of using the hash in the filename, but with the `/` all replaced with `$` so that there's no issue with certain filenames. We can use AES if we think that the hashes being too similar is a problem (but I'm also not sure how to use the `aes` crate).

PR #295, opened 2020-01-22
Run Fixing Algorithms when opening splits

### LINKED ISSUE #21: Run Fixing Algorithms when opening splits

TheEssenceOfDarkness:
closes #21

edit: I'm probably missing something here. All of the tests are still passing on my machine though.

### DIFF
--- a/src/run/editor/mod.rs
+++ b/src/run/editor/mod.rs
@@ -80,7 +80,8 @@ pub struct Editor {
 impl Editor {
     /// Creates a new Run Editor that modifies the Run provided. Creation of the
     /// Run Editor fails when a Run with no segments is provided.
-    pub fn new(run: Run) -> Result<Self, OpenError> {
+    pub fn new(mut run: Run) -> Result<Self, OpenError> {
+        run.fix_splits();
         let len = run.len();
         if len == 0 {
             return Err(OpenError::EmptyRun);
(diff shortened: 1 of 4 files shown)


CryZe:
You may not have run `cargo update` for a while, so some dependencies are slightly outdated, which is why the tests still work for you locally. You may want to rebase your branch onto the latest master anyway, where we use the perceptual hashing now.

Also you should probably add a test here where we check that the Run does indeed get fixed when creating a Timer / Run Editor / ...

src/timing/timer/tests/mod.rs
```diff
@@ -617,3 +617,51 @@ fn reset_and_set_attempt_as_pb() {
...
+    let run = timer.into_run(true);
+
+    let segments = run.segments();
+
+    assert_eq!(segments[0].segment_history().try_get_min_index(), Some(0));
+    assert_eq!(segments[0].segment_history().try_get_max_index(), Some(1));
+
+    assert_eq!(segments[1].segment_history().try_get_min_index(), Some(0));
+    assert_eq!(segments[1].segment_history().try_get_max_index(), Some(1));
+}
```
TheEssenceOfDarkness: These tests will not compile due to run.attempt_history being private.

Should I move these tests to `crate::run::tests`, or is there something else I can do here?

CryZe: https://docs.rs/livesplit-core/0.11.0/livesplit_core/run/struct.Run.html#method.attempt_history

CryZe: Oh wait, you try to modify it. Uhm, then yeah I guess move it there

TheEssenceOfDarkness:
This should be good now.

PR #303, opened 2020-02-07
Update image to 0.23.0

CryZe:
This also finally gets rid of 0.x versions of syn, which is a slow dependency to compile.

This is blocked by https://github.com/abonander/img_hash/pull/33 being merged.

### DIFF
--- a/src/settings/image/shrinking.rs
+++ b/src/settings/image/shrinking.rs
@@ -38,10 +38,10 @@ fn shrink_inner(data: &[u8], max_dim: u32) -> Result<Cow<'_, [u8]>, ImageError> 
             DynamicImage::ImageRgba8(x) => x.as_ref(),
             DynamicImage::ImageBgr8(x) => x.as_ref(),
             DynamicImage::ImageBgra8(x) => x.as_ref(),
-            DynamicImage::ImageLuma16(x) => unsafe { x.align_to::<u8>().1 },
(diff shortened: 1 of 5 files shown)


src/settings/image/shrinking.rs
```diff
@@ -10,34 +10,46 @@ fn shrink_inner(data: &[u8], max_dim: u32) -> Result<Cow<'_, [u8]>, ImageError>
...
+            DynamicImage::ImageLuma8(x) => x.as_ref(),
+            DynamicImage::ImageLumaA8(x) => x.as_ref(),
+            DynamicImage::ImageRgb8(x) => x.as_ref(),
+            DynamicImage::ImageRgba8(x) => x.as_ref(),
+            DynamicImage::ImageBgr8(x) => x.as_ref(),
+            DynamicImage::ImageBgra8(x) => x.as_ref(),
+            DynamicImage::ImageLuma16(x) => unsafe { x.align_to::<u8>().1 },
+            DynamicImage::ImageLumaA16(x) => unsafe { x.align_to::<u8>().1 },
+            DynamicImage::ImageRgb16(x) => unsafe { x.align_to::<u8>().1 },
+            DynamicImage::ImageRgba16(x) => unsafe { x.align_to::<u8>().1 },
```
CryZe: This isn't quite right as `align_to` isn't actually required to actually do anything. Especially on miri it just doesn't.

PR #307, opened 2020-03-01
Move the Title Abbreviations into the Component

CryZe:
So far the renderer took care of handling the abbreviations for the title, but other frontends may want to make use of them as well, such is the case in the React based renderer in the web frontend of LiveSplit One.

### DIFF
--- a/src/component/title/mod.rs
+++ b/src/component/title/mod.rs
@@ -9,7 +9,7 @@ use crate::{
     settings::{CachedImageId, Image, ImageData},
     Timer, TimerPhase,
 };
-use alloc::borrow::Cow;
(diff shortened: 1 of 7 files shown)


CryZe:
Welp, this is what I was fearing. They just removed OpenSSL in cross, so we can't cross compile to "obscure" targets with networking active anymore.

src/component/title/mod.rs
```diff
@@ -307,3 +351,51 @@ impl Component {
         }
     }
 }
+
+// FIXME: Turn this into a generator once that works on stable Rust.
+fn abbreviate_category(category: Cow<'_, str>) -> Vec<Box<str>> {
```
wooferzfg: Are you sure that this method should belong here instead of expanding the scope of `livesplit-title-abbreviations` and putting it there?

PR #310, opened 2020-03-07
Implement Gamepad Support for the Web

CryZe:
This adds support for using gamepad buttons as hotkeys. For now this is only implemented for the web.


crates/livesplit-hotkey/src/wasm_web/mod.rs
```diff
@@ -52,10 +55,84 @@ impl Hook {
...
+                        }
+                    }
+                }
+            }
+        }) as Box<dyn FnMut()>);
+
+        let interval_id = window
+            .set_interval_with_callback_and_timeout_and_arguments_0(
+                gamepad_callback.as_ref().unchecked_ref(),
+                1000 / 60,
```
wooferzfg: Should the refresh rate (60 fps) be extracted as a constant?

CryZe: Considering it's not used anywhere else, I don't think so. Constants for the sake of constants is not useful imo.

PR #315, opened 2020-03-12
Fix the Single Line Title Calculation

### LINKED ISSUE #314: Titles in the layout state of an Original Livesplit layout are no longer calculated correctly.
After switching from the crates.io build (livesplit-core = "0.11.0") to using the master branch in git, the title of a run is no longer properly calculated when calling Layout.state on a layout from the original livesplit with **SingleLine=True**.

These are the layouts used for this report.
[Layouts.zip](https://github.com/LiveSplit/livesplit-core/files/4324826/Layouts.zip)

## Screenshots
Taken in one.livesplit.org
Both use the same run
LiveSplitCore Layout Titles

## To Reproduce
Assume **old_layout** is a previously loaded **Layout** from the **old_layout.lsl** file (This layout was created on the original LiveSplit)
Assume **new_layout** is a previously loaded **Layout** from the **new_layout.ls1l** file (This layout was created on one.livesplit.org)
(Both layouts have **SingleLine=True**)
Assume **timer** is a previously initialized **SharedTimer** that contains a run with a game name and a category name.
```
let old_layout_state = old_layout.state(&timer.read());
let new_layout_state = new_layout.state(&timer.read());
for component in &old_layout_state.components {
    match component {
        ComponentState::Title(old_state) => {
			for component in &new_layout_state.components {
				match component {
					ComponentState::Title(new_state) => {
						if old_state.line1.len() > 0 && new_state.line1.len() > 0 {
							let old_line1 = &old_state.line1[old_state.line1.len() - 1];
							let old_line1 = &old_line1[..];
							let new_line1 = &new_state.line1[new_state.line1.len() - 1];
							let new_line1 = &new_line1[..];
							assert_eq!(old_line1, new_line1);
						}
						else {
							assert_eq!(old_state.line1.len(), new_state.line1.len());
						}
					}
				}
			}
        },
        _ => (),
    }
}
```

## Notes
This issue did not happen when using crates.io build (livesplit-core = "0.11.0")

CryZe:
It assumed that for a single line that the game and category always exist, but that is not always the case. We already handled that case, just not for the single line title scenario. So this basically moves the whole single line title calculation into the branch where we made sure that both exist in the first place.

Resolves #314

### DIFF
--- a/src/component/title/mod.rs
+++ b/src/component/title/mod.rs
@@ -211,64 +211,66 @@ impl Component {
             self.settings.show_variables,
         );
 
-        let (line1, line2) = if self.settings.display_as_single_line {
-            let mut abbrevs = Vec::new();
-            let mut abbrev = String::new();
-
-            if !full_category_name.is_empty() {
-                for game_abbrev in game_abbrevs.iter() {
-                    abbrev.clear();
-                    abbrev.push_str(game_abbrev);
-                    if !game_abbrev.is_empty() {
-                        abbrev.push_str(" - ");
-                    }
-                    abbrev.push_str(&full_category_name);
-                    abbrevs.push(abbrev.as_str().into());
-                }
-            }
-            // This assumes the last element is the unabbreviated value, which
-            // can only be the case if the `game_abbrevs` also has the
-            // unabbreviated game name as its last element.
-            let swap_index = abbrevs.len().checked_sub(1);
-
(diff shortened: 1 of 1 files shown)


Ynscription:
After further investigation it turns out that it had nothing to do with the single line, the problem comes from the layout:
LiveSplitCore Layout Title

PR #316, opened 2020-03-13
Various Improvements to the C API

CryZe:
1. A `get_buf_len` function is introduced that helps languages skip the string length calculation. This should reduce some overhead.
2. Adds a way to allow JavaScript to obtain the splits as an `Uint8Array`. This is super cheap as it directly points into the wasm memory without allocating anything on JavaScript's side.
3. Turns the `Int8Array` in the JavaScript bindings to `Uint8Array`. This more closely reflects the `&[u8]` / `Vec<u8>` on Rust's side.

### DIFF
--- a/capi/bind_gen/src/wasm_bindgen.rs
+++ b/capi/bind_gen/src/wasm_bindgen.rs
@@ -460,7 +460,7 @@ interface Slice {
     len: number,
 }
 
-function allocInt8Array(src: Int8Array): Slice {
+function allocUint8Array(src: Uint8Array): Slice {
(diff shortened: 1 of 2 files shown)


DarkRTA:
We should probably find a way to namespace `get_buf_len` and maybe even future future functions that are named similarly.

Unlike the naming of our other C-API functions this one is actually pretty likely to cause a conflict somewhere.

CryZe:
Oh yeah, I didn't think of that. I guess we'll change the name at some point.

PR #318, opened 2020-03-28
Properly read strings as UTF-8 in C# bindings

### LINKED ISSUE #120: C# doesn't seem to marshal its strings as UTF-8
https://github.com/LiveSplit/LiveSplit/issues/1054 seems to imply that the strings are actually marshalled as Shift-JIS on japanese Windows. So we probably should marshal them differently. The FFI Omnibus should be fixed accordingly too: https://github.com/shepmaster/rust-ffi-omnibus/issues/63

CryZe:
The encoding previously was locale dependent, so for some languages such as Japanese the default encoding is not UTF-8. We now explicitly do the UTF-8 decoding, so this should work across all setups now.

Additionally the string length is known by Rust, so we don't need to do a loop looking for the nul-terminator anymore.

Resolves #120

### DIFF
--- a/capi/bind_gen/src/csharp.rs
+++ b/capi/bind_gen/src/csharp.rs
@@ -15,19 +15,18 @@ fn get_hl_type(ty: &Type) -> String {
     } else if ty.name == "usize" {
         "long".to_string()
     } else {
-        get_ll_type(ty, false).to_string()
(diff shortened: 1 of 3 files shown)


CryZe:
@wooferzfg This is entirely untested. We should probably do a PR in LiveSplit as well that updates the DLLs and bindings.

PR #321, opened 2020-04-11
Keep the animation of the GIFs when shrinking them

CryZe:
Previously we shrank each image down to a PNG every time, but PNGs can not have animations, so animations for GIFs were discarded. Now we shrink GIFs down to GIFs and retain their animations.


CryZe:
Theoretically we could also check if the GIF even has more than one frame and if it doesn't shrink it to a PNG.

wooferzfg:
This is doing something weird.

Original GIF:

tumblr_nhu67aaimV1r7sijxo1_400

Scaled GIF in LiveSplit:

xTLtV21hEC

CryZe:
Can you send me both gifs? (possibly as "raw files" so github / imgur / ... don't mess with it)

PR #330, opened 2020-05-10
Expose the Skill Curve in the API

CryZe:
We've used the Balanced PB algorithm for various things by now: Balanced PB itself, the goal generator, the PB chance and it theoretically could also be used for the median segments and for calculating the simple sum of best and worst segments. The underlying curve that is being sampled here contains lots of interesting information, so it makes sense to expose it for analysis purposes in the API. I dubbed it the "skill curve" here, but the name is not necessarily final.

### DIFF
--- a/src/analysis/pb_chance/mod.rs
+++ b/src/analysis/pb_chance/mod.rs
@@ -5,8 +5,9 @@
 //! result as the PB chance for the Run. The value is being reported as a
 //! floating point number in the range from 0 (0%) to 1 (100%).
 //!
-//! The PB chance is currently calculated through the Balanced PB algorithm. The
-//! PB chance is the percentile at which the Balanced PB algorithm finds the PB.
+//! The PB chance is currently calculated with the skill curve. The PB chance is
+//! the percentile at which the PB is located on the skill curve. This is also
+//! where the Balanced PB would source its split times.
 
 use super::SkillCurve;
 use crate::{comparison, Run, Segment, TimeSpan, Timer, TimingMethod};
--- a/src/analysis/pb_chance/mod.rs
+++ b/src/analysis/pb_chance/mod.rs
@@ -8,7 +8,7 @@
 //! The PB chance is currently calculated through the Balanced PB algorithm. The
 //! PB chance is the percentile at which the Balanced PB algorithm finds the PB.
 
-use crate::platform::prelude::*;
+use super::SkillCurve;
 use crate::{comparison, Run, Segment, TimeSpan, Timer, TimingMethod};
 
 #[cfg(test)]
@@ -24,17 +24,7 @@ fn calculate(segments: &[Segment], method: TimingMethod, offset: TimeSpan) -> f6
         return 1.0;
     }
 
-    let mut all_weighted_segment_times = vec![Vec::new(); segments.len()];
-    let mut time_span_buf = Vec::with_capacity(segments.len());
-
-    comparison::goal::determine_percentile(
-        offset,
-        segments,
-        method,
-        None,
-        &mut time_span_buf,
-        &mut all_weighted_segment_times,
-    )
+    comparison::goal::determine_percentile(offset, segments, method, None, &mut SkillCurve::new())
 }
 
 /// Calculates the PB chance for a run. No information about an active attempt
--- a/src/analysis/skill_curve.rs
+++ b/src/analysis/skill_curve.rs
@@ -7,7 +7,7 @@ const WEIGHT: f64 = 0.75;
 const TRIES: usize = 50;
 
 /// The skill curve analyzes the segment history across all segments. For each
-/// segment all the segment times are sorted by length and weighted by their
+/// segment, all the segment times are sorted by length and weighted by their
 /// recency. Plotting this on a graph with the y-axis representing the segment
 /// time and the x-axis representing the percentile, with the shortest time at 0
 /// and the longest time at 1, yields the so called "skill curve". If you sum
@@ -16,9 +16,9 @@ const TRIES: usize = 50;
 ///
 /// # Properties of the Skill Curve
 ///
-/// If you sample the curve at 0, you get the simple sum of best segments and if
-/// you sample the curve at 1, you get the simple sum of worst segments. At 0.5
-/// you get the median segments. If you sample the individual segments where you
+/// If you sample the curve at 0, you get the simple sum of best segments, and if
+/// you sample the curve at 1, you get the simple sum of worst segments. At 0.5,
+/// you get the median segments. If you sample the individual segments at the same percentile where you
 /// find the Personal Best on the overall run's curve, you get the Balanced PB.
 /// The position of the Balanced PB on the x-axis is the PB chance.
 #[derive(Default, Clone)]
@@ -33,7 +33,7 @@ impl SkillCurve {
         Default::default()
     }
 
-    /// Returns the amount of segments this skill curve is comprised of.
+    /// Returns the number of segments this skill curve is comprised of.
     pub fn len(&self) -> usize {
         self.all_weighted_segment_times.len()
     }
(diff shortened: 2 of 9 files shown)


CryZe:
I've also tried interpolation search, which should search the curve much faster (as a separate local commit), but the curve usually has a fairly exponential shape (it's much rarer to play extremely well than extremely bad), so the interpolation search is constantly underestimating where the time would be. We would need to do a variant of interpolation search that doesn't just linearly interpolate.

src/analysis/skill_curve.rs
```diff
@@ -0,0 +1,223 @@
...
+/// segment all the segment times are sorted by length and weighted by their
+/// recency. Plotting this on a graph with the y-axis representing the segment
+/// time and the x-axis representing the percentile, with the shortest time at 0
+/// and the longest time at 1, yields the so called "skill curve". If you sum
+/// all the different curves together for all the segments, you get the overall
+/// curve for the whole run.
+///
+/// # Properties of the Skill Curve
+///
+/// If you sample the curve at 0, you get the simple sum of best segments and if
```
wooferzfg, line 19:
```suggestion
/// If you sample the curve at 0, you get the simple sum of best segments, and if
```

src/analysis/skill_curve.rs
```diff
@@ -0,0 +1,223 @@
...
+/// you sample the curve at 1, you get the simple sum of worst segments. At 0.5
```
wooferzfg, line 20:
```suggestion
/// you sample the curve at 1, you get the simple sum of worst segments. At 0.5,
```

src/analysis/skill_curve.rs
```diff
@@ -0,0 +1,223 @@
...
+/// you get the median segments. If you sample the individual segments where you
```
wooferzfg, line 21:
```suggestion
/// you get the median segments. If you sample the individual segments at the same percentile where you
```

src/analysis/skill_curve.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+impl SkillCurve {
+    /// Constructs a new empty skill curve. Before querying information, you
+    /// need to calculate the curve for some segments.
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// Returns the amount of segments this skill curve is comprised of.
```
wooferzfg, line 36:
```suggestion
    /// Returns the number of segments this skill curve is comprised of.
```

src/analysis/skill_curve.rs
```diff
@@ -0,0 +1,223 @@
...
+    pub fn len(&self) -> usize {
+        self.all_weighted_segment_times.len()
+    }
+
+    /// Returns `true` if there are no segments in this skill curve.
+    pub fn is_empty(&self) -> bool {
+        self.all_weighted_segment_times.is_empty()
+    }
+
+    /// Reduces the amount of segments that are being considered by this curve.
```
wooferzfg, line 46:
```suggestion
    /// Reduces the number of segments that are being considered by this curve.
```

src/analysis/skill_curve.rs
```diff
@@ -0,0 +1,223 @@
...
+                match found_index {
+                    // The percentile perfectly matched a segment time
+                    Ok(index) => weighted_segment_times[index].1,
+                    // The percentile didn't perfectly match, interpolate instead
+                    Err(right_index) => {
+                        // Saturate both left and right index at the bounds of the
+                        // array. Both could go out of bounds.
+                        let left_index = right_index.saturating_sub(1);
+                        // This assumes len can never be 0. This only works out
+                        // for us, because we truncate all the empty segments away.
```
wooferzfg, line 150:
```suggestion
                        // for us because we truncate all the empty segments away.
```

src/analysis/skill_curve.rs
```diff
@@ -0,0 +1,223 @@
...
+        self.iter_segment_times_at_percentile(percentile)
+            .map(move |segment_time| {
+                sum += segment_time;
+                sum
+            })
+    }
+
+    /// Searches the curve for the final run time specified and returns the
+    /// percentile the time can be found at. The percentile is always within the
+    /// range 0..1. If the time specified can not be found, the percentile
```
wooferzfg, line 187:
```suggestion
    /// range 0..1. If the time specified cannot be found, the percentile
```

src/analysis/pb_chance/mod.rs
```diff
@@ -8,7 +8,7 @@
 //! The PB chance is currently calculated through the Balanced PB algorithm. The
 //! PB chance is the percentile at which the Balanced PB algorithm finds the PB.
```
wooferzfg, line 9: I think we should update this explanation to mention skill curves instead of "the Balanced PB algorithm".

src/analysis/skill_curve.rs
```diff
@@ -0,0 +1,223 @@
...
+            let diff = max - min;
+
+            if diff != 0.0 {
+                for (weight, _) in weighted_segment_times.iter_mut() {
+                    *weight = (*weight - min) / diff;
+                }
+            }
+        }
+
+        // Limit the slice to only the segments that have segment times.
```
wooferzfg, line 123: since the other comments don't end with a period:
```suggestion
        // Limit the slice to only the segments that have segment times
```

src/analysis/skill_curve.rs
```diff
@@ -0,0 +1,223 @@
...
+            .iter()
+            .map(move |weighted_segment_times| {
+                let found_index = weighted_segment_times
+                    .binary_search_by(|&(w, _)| w.partial_cmp(&percentile).unwrap());
+
+                match found_index {
+                    // The percentile perfectly matched a segment time
+                    Ok(index) => weighted_segment_times[index].1,
+                    // The percentile didn't perfectly match, interpolate instead
+                    Err(right_index) => {
```
wooferzfg, line 145: Should we mention in the documentation for this function that you get an `Err` for any segment where you don't have an exact percentile match?

src/analysis/skill_curve.rs
```diff
@@ -0,0 +1,223 @@
...
+            // Binary search the correct percentile
+            match sum.partial_cmp(&time_to_find) {
+                Some(Ordering::Equal) => return percentile,
+                Some(Ordering::Less) => perc_min = percentile,
+                Some(Ordering::Greater) => perc_max = percentile,
+                None => {}
+            }
+        }
+
+        (perc_max + perc_min) / 2.0
```
wooferzfg, line 209: Should we be distinguishing between exact and inexact matches here like we do for `iter_segment_times_at_percentile`?

wooferzfg:
A comment about the stats here - the actual chance of a PB should be much lower than the "PB chance" that we calculate if each segment's curve is exponential. We're doing our calculation based on the assumption that you'll play equally well across all segments, but that's not actually what happens in practice. I think this is ok though - I'd rather overestimate the PB chance in some cases rather than showing a discouraging number.

src/analysis/skill_curve.rs
CryZe: You don't get an error. The curve is a linear interpolation of the individual segment times you got. Whether we need to interpolate or not is figured out by the binary search that is done. If it doesn't exactly find a match we have to interpolate. But that's almost always the case.

src/analysis/skill_curve.rs
CryZe: Here we could signal to the outside that it's not an exact match, but idk if we really need one. 50 tries of binary search yields an accuracy of 50 bits, which is close to the 64-bits of accuracy f64 could give us. So if we really cared for full accuracy, we could just do 64 rounds.

src/analysis/skill_curve.rs
CryZe: Ok / Err is just the Rust standard library's way to signal the result of the binary search. https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search

wooferzfg: That makes sense, thanks for the clarification.

PR #334, opened 2020-05-30
Implement updating the Layout States

### LINKED ISSUE #54: Reuse allocations by reusing state objects.
~~We should reduce unnecessary allocations as much as possible. I wasn't all too sure if Cow plays well with serde, so I didn't use it until now, but I just tested it and it works to seem fine, so yeah, we should use it.~~

CryZe:
So far we've always discarded the layout states, but since we want to calculate layout states every frame, it makes a lot more sense to keep the layout state around and just update it. All of the original API is retained, so you are not forced to actually update your layout states. However the code for creating a new layout state now initializes a default state and updates it, instead of directly creating it correctly. This might slightly slow down initial creation, but we don't want to basically duplicate the same code across initial creation and updating.

By doing this an average frame now has 0 allocations, reallocations and deallocations. We are seeing speedups of around 3x - 5x in the newly added benchmarks.

This also introduces a new `ClearVec` type that behaves like a normal `Vec` but when clearing it, instead of dropping all the elements, it just keeps them around and clears them instead. So if you had a `Vec<String>` and cleared it, all the Strings would get deallocated, while a `ClearVec<String>` would keep them around and just set their length back to 0. We use this in a few places to reduce the chances of us needing to allocate new memory. It might make sense to move it into its own crate eventually as this should be a useful type for any kind of Rust code.

Resolves #54

### DIFF
--- a/capi/bind_gen/src/typescript.ts
+++ b/capi/bind_gen/src/typescript.ts
@@ -137,11 +137,11 @@ export interface TitleComponentStateJson {
     /**
      * By default the category name is shown on the second line. Based on the
      * settings, it can however instead be shown in a single line together with
-     * the game name. This is a list of all the possible abbreviations. It
-     * contains at least one element and the last element is the unabbreviated
-     * value.
+     * the game name. This is a list of all the possible abbreviations. If this
+     * is empty, only a single line is supposed to be shown. If it contains at
+     * least one element, the last element is the unabbreviated value.
      */
-    line2: string[] | null,
+    line2: string[],
     /**
      * Specifies whether the title should centered or aligned to the left
      * instead.
--- a/capi/src/layout.rs
+++ b/capi/src/layout.rs
@@ -4,7 +4,7 @@
 use super::{get_file, output_vec, release_file, str, Json};
 use crate::component::OwnedComponent;
 use crate::layout_state::OwnedLayoutState;
-use livesplit_core::layout::{parser, LayoutSettings};
+use livesplit_core::layout::{parser, LayoutSettings, LayoutState};
 use livesplit_core::{Layout, Timer};
 use std::io::{BufReader, Cursor};
 use std::slice;
@@ -91,6 +91,25 @@ pub unsafe extern "C" fn Layout_parse_original_livesplit(
 #[no_mangle]
 pub extern "C" fn Layout_state(this: &mut Layout, timer: &Timer) -> OwnedLayoutState {
     Box::new(this.state(timer))
+}
+
+/// Updates the layout's state based on the timer provided.
+#[no_mangle]
+pub extern "C" fn Layout_update_state(this: &mut Layout, state: &mut LayoutState, timer: &Timer) {
+    this.update_state(state, timer)
+}
+
+/// Updates the layout's state based on the timer provided.
+#[no_mangle]
+pub extern "C" fn Layout_update_state_as_json(
+    this: &mut Layout,
+    state: &mut LayoutState,
+    timer: &Timer,
+) -> Json {
+    this.update_state(state, timer);
+    output_vec(|o| {
+        state.write_json(o).unwrap();
+    })
 }
 
 /// Calculates the layout's state based on the timer provided and encodes it as
--- a/capi/src/layout.rs
+++ b/capi/src/layout.rs
@@ -99,7 +99,8 @@ pub extern "C" fn Layout_update_state(this: &mut Layout, state: &mut LayoutState
     this.update_state(state, timer)
 }
 
-/// Updates the layout's state based on the timer provided.
+/// Updates the layout's state based on the timer provided and encodes it as
+/// JSON.
 #[no_mangle]
 pub extern "C" fn Layout_update_state_as_json(
     this: &mut Layout,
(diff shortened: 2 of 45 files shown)


capi/src/layout.rs
```diff
@@ -93,6 +93,25 @@ pub extern "C" fn Layout_state(this: &mut Layout, timer: &Timer) -> OwnedLayoutS
     Box::new(this.state(timer))
 }
 
+/// Updates the layout's state based on the timer provided.
+#[no_mangle]
+pub extern "C" fn Layout_update_state(this: &mut Layout, state: &mut LayoutState, timer: &Timer) {
+    this.update_state(state, timer)
+}
+
+/// Updates the layout's state based on the timer provided.
```
wooferzfg, line 102: Should this also say `and encodes it as JSON`?

Also, it might be useful to explain the difference somewhere between the use case for `Layout_update_state_as_json` vs. `Layout_state_as_json`.

capi/src/title_component_state.rs
```diff
@@ -46,8 +46,8 @@ pub extern "C" fn TitleComponentState_line1(this: &TitleComponentState) -> *cons
 pub extern "C" fn TitleComponentState_line2(this: &TitleComponentState) -> *const Nullablec_char {
     // FIXME: Add API for querying the abbreviations.
     this.line2
-        .as_ref()
-        .map_or_else(ptr::null, |abbrevs| output_str(&abbrevs.last().unwrap()))
+        .last()
+        .map_or_else(ptr::null, |abbrev| output_str(&abbrev))
```
wooferzfg, line 50: Is it ok that we can return null here but not in the TypeScript bindings?

src/clear_vec.rs
```diff
@@ -0,0 +1,212 @@
+//! A ClearVec is a special kind of Vec that when being cleared doesn't drop its
+//! elements. Instead the elements get cleared out. This allows reusing them
+//! later on again with all their capacity retained. The elements are only
+//! dropped once the ClearVec itself gets dropped.
```
wooferzfg, line 4:
```suggestion
//! A ClearVec is a special kind of Vec that doesn't drop its elements when
//! being cleared. Instead, the elements get cleared out. This allows reusing
//! them later on again with all their capacity retained. The elements are only
//! dropped once the ClearVec itself gets dropped.
```

src/clear_vec.rs
```diff
@@ -0,0 +1,212 @@
...
+/// that's not a hard requirement.
+pub trait Clear {
+    /// Clears the object.
+    fn clear(&mut self);
+}
+
+/// A ClearVec is a special kind of Vec that when being cleared doesn't drop its
+/// elements. Instead the elements get cleared out. This allows reusing them
+/// later on again with all their capacity retained. The elements are only
+/// dropped once the ClearVec itself gets dropped.
```
wooferzfg, line 29:
```suggestion
/// A ClearVec is a special kind of Vec that doesn't drop its elements when
/// being cleared. Instead, the elements get cleared out. This allows reusing
/// them later on again with all their capacity retained. The elements are only
/// dropped once the ClearVec itself gets dropped.
```

src/comparison/mod.rs
```diff
@@ -118,8 +116,17 @@ pub fn shorten(comparison: &str) -> &str {
...
-    }
+        _ => return Option::None,
+    })
+}
+
+/// Shortens a comparison name. If the name of the comparison matches one of the
+/// comparison generators, the short name of that comparison generator is
+/// returned. Otherwise the comparison name is returned without being shortened.
+/// Additional shortening logic for other comparison names may happen in the
+/// future.
```
wooferzfg, line 127:
```suggestion
/// returned. Otherwise, the comparison name is returned without being
/// shortened. Additional shortening logic for other comparison names may happen
/// in the future.
```

capi/bind_gen/src/typescript.ts
```diff
@@ -137,11 +137,11 @@ export interface TitleComponentStateJson {
...
      * settings, it can however instead be shown in a single line together with
-     * the game name. This is a list of all the possible abbreviations. It
-     * contains at least one element and the last element is the unabbreviated
-     * value.
+     * the game name. This is a list of all the possible abbreviations. If this
+     * is empty, only a single line is supposed to be shown. If it contains at
+     * least one element, the last element is the unabbreviated value.
      */
-    line2: string[] | null,
+    line2: string[],
```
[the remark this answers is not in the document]
CryZe: Yeah

capi/src/layout.rs
```diff
@@ -93,6 +93,25 @@ pub extern "C" fn Layout_state(this: &mut Layout, timer: &Timer) -> OwnedLayoutS
...
+#[no_mangle]
+pub extern "C" fn Layout_update_state_as_json(
```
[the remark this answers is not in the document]
CryZe: That's what I've been using in my experimental branch. I'm not sure if we'll keep this function for too long, but for now it is the faster function.

capi/src/title_component_state.rs
CryZe: We don't use this function at all, since we serialize and deserialize the state to JSON. This is the alternative way if you don't serialize. This is what @DarkRTA is calling in his timer. The function here has no notion of abbreviations at all (because it was never introduced in this part of the C API), so it basically still only has an optional second line as a single string.

PR #339, opened 2020-06-10
Introduce a Timer Snapshot Mechanism

CryZe:
This introduces the ability to create snapshots of the timer. At the moment of snapshot creation the current time of the timer is evaluated. The snapshot then represents the timer with the current time frozen at that specific time. This allows layout state evaluation and other calculations to work with a consistent timer that is entirely frozen, allowing for more consistent results. However almost more importantly, this reduces the amount of times we need to take time stamps from the underlying system, improving the performance a bit.

In the web the performance improvement is even more significant. There the layout state calculation improved from 8.5us to 5us.

### DIFF
--- a/src/timing/timer/mod.rs
+++ b/src/timing/timer/mod.rs
@@ -2,7 +2,7 @@ use crate::comparison::personal_best;
 use crate::platform::prelude::*;
 use crate::TimerPhase::*;
 use crate::{AtomicDateTime, Run, Segment, Time, TimeSpan, TimeStamp, TimerPhase, TimingMethod};
-use core::mem;
+use core::{mem, ops::Deref};
 
 #[cfg(test)]
 mod tests;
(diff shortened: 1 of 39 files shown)


src/timing/timer/mod.rs
```diff
@@ -197,6 +218,17 @@ impl Timer {
             .with_game_time(game_time)
     }
 
+    /// Creates a new snapshot of the timer at the point of time of this call.
```
wooferzfg, line 221:
```suggestion
    /// Creates a new snapshot of the timer at the point in time of this call.
```

CryZe:
The Synchronization Protocol and the Time Sources are mostly concerned about how you modify / control the timer. We need some low level mechanism that does indeed affect https://github.com/LiveSplit/livesplit-core/blob/706361a8289df8793e826b13110cb34bb9af5089/src/timing/timer/mod.rs#L317

For perceiving the timer however the snapshot mechanism is what we want. I don't think this clashes with the design of manipulating the timer.

PR #340, opened 2020-06-13
Delay registering the Gamepad Hook

CryZe:
In the web we support gamepads. Unfortunately there's no callbacks or so that we could use, so instead we have to use an interval where we poll for button presses. This isn't super expensive, but still unnecessary if you don't use a gamepad. Additionally Chrome has the problem that it gets exclusive access to the gamepads and doesn't really yield control back until the whole browser is closed. So you really don't want that unless you really intend to use the gamepad for splitting. This is why the gamepad interval is now only being set when the first gamepad button is registered.

### DIFF
--- a/crates/livesplit-hotkey/src/wasm_web/mod.rs
+++ b/crates/livesplit-hotkey/src/wasm_web/mod.rs
@@ -5,8 +5,11 @@ use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
 use web_sys::{window, Gamepad, GamepadButton, KeyboardEvent};
 
-use std::collections::hash_map::{Entry, HashMap};
-use std::sync::{Arc, Mutex};
+use std::{
+    cell::Cell,
+    collections::hash_map::{Entry, HashMap},
+    sync::{Arc, Mutex},
+};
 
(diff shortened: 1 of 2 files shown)


crates/livesplit-hotkey/src/wasm_web/mod.rs
```diff
@@ -31,11 +34,37 @@ impl Drop for Hook {
                 "keypress",
                 self.keyboard_callback.as_ref().unchecked_ref(),
             );
-            window.clear_interval_with_handle(self.interval_id);
+            if let Some(interval_id) = self.interval_id.get() {
+                window.clear_interval_with_handle(interval_id);
```
wooferzfg, line 38: It's definitely an edge case, but should we also be calling this in `unregister` if there's no more gamepad buttons registered anymore?

CryZe: I don't think that case matters much, as I think it's a super rare situation that someone was using gamepad, then stops using gamepad. And even then, the cost for the gamepad hook is still very minimal, and it stops being a cost the next time they reload. So overall, I'd say yes, we could possibly remove the hook when unregistering all gamepad buttons, but it has a much lesser impact.

PR #344, opened 2020-06-25
Add dependabot.yml

wooferzfg:
Setting up updates for GitHub Actions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: daily
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: daily
```
CryZe, line 6: Does this override the behavior we already set up for crates auto updates?

wooferzfg: We currently get updates from dependabot-preview, but this config file enables GitHub's dependabot. So I think we'd have to disable dependabot-preview and also make sure that this config file matches the behavior that we have in the dependabot dashboard.

wooferzfg: Actually, I'm not sure how you'd disable dependabot-preview (or if you even have to do that). I'll look into that.

wooferzfg: I looked into this a bit more. [This post](https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/) makes it sound like creating the config file is all you need to do in order to migrate from dependabot-preview to the new dependabot. However, we'll need to update this PR to match our existing settings on the dependabot.com dashboard before merging this.

wooferzfg:
New Dependabot doesn't support auto-merge yet, so I'm closing this.

PR #355, opened 2020-09-09
Add support for Auto Splitting to the c api. 

DarkRTA:
This works in its current state but some stuff needs to be adjusted.

-----
### Todo:

- [ ] Find a way to expose logging as right now logging just does not work.

- [x] Find a way to make auto splitting support optional.

- [x] ~~Improve bindings generator to conditionally include definitions for the auto splitting feature.~~ Stubbed out functions instead.


capi/Cargo.toml
```diff
@@ -5,7 +5,7 @@ authors = ["Christopher Serr <christopher.serr@gmail.com>"]
 edition = "2018"
 
 [dependencies]
-livesplit-core = { path = "..", default-features = false, features = ["std"] }
+livesplit-core = { path = "..", default-features = false, features = ["std", "auto-splitting"] }
```
CryZe, line 8: This is the reason why I didn't pull it in yet. The problem is that unconditionally depending on auto splitting will break for example the web version. So this needs to be an optional feature. However at least atm we don't support features in the binding generator. So either the C API needs to stub out the implementation or we need to improve our binding generator.

CryZe (changes requested):

capi/Cargo.toml
DarkRTA: I've added the feature attribute to everything now, but I still need to work on improving the bindings generator.

That might take me a while since I'm not familiar at all with `syn` and binding generator's code is a bit messy.

DarkRTA:
The only thing blocking this right now is the fact that we yet to improve the bindings generator. (All of the needed feature attributes have been set up.)

I've tried doing it myself but `syn` is being a pain in the ass to deal with and I cant seem to get it to do what I want. If anyone has any guidance on this I'd really appreciate it.

DarkRTA:
I'm going to opt to just stub out the implementation here instead of extending the bindings generator, using [this](https://github.com/CryZe/livesplit-core/blob/a7c7e0d2f8edca18ec27ce560be3b97c1acc6a0a/capi/src/software_renderer.rs#L6-L15) as an example.

We can improve the bindings generator at a later date.

PR #356, opened 2020-09-10
Add a Linux-only tool needed for Auto Splitting to work correctly.

DarkRTA:
This adds a crate called `livesplit-exec` which is a small tool needed for Auto Splitting to work correctly on Linux.

A script called `build.sh` is provided for convenience but I can remove this if desired.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,11 @@ vek = { version = "0.10.1", default-features = false, features = ["libm"], optio
 # Networking
 splits-io-api = { version = "0.1.2", optional = true }
 
+# Auto Splitting
+livesplit-auto-splitting = { path = "crates/livesplit-auto-splitting", version = "0.1.0", optional = true }
+crossbeam-channel = { version = "0.3.6", default-features = false, optional = true }
+log = { version = "0.4.6", optional = true }
+
 [target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
 # WebAssembly in the Web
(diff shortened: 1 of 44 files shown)


CryZe:
I'd like to see a significant amount of documentation why this is necessary, why there's no other option and how someone using auto splitting on Linux needs to use this. Also keep in mind that the auto splitter runtime is meant to be a library that can be used outside of LiveSplit One / livesplit-core, so a general explanation is necessary. Preferably all of that would be in the README that you already created.

DarkRTA:
I went ahead and wrote some more documentation for this. Let me know if there's anything else you want changed.

kitlith:
Could you add a line or two to the documentation about how this is only necessary as a workaround for the mitigations introduced with https://www.kernel.org/doc/html/v4.15/admin-guide/LSM/Yama.html ? Does this work with mode 2 or just mode 1?

DarkRTA:
@kitlith: I've pretty much rewrote the first paragraph to explicity mention Yama and added a note about it not working if `ptrace_scope > 1`. I did oversimplify it a bit for terseness but its enough to get why its needed.

@CryZe: If you and kitlith feel this is sufficient, feel free to merge.

kitlith:
LGTM.

I will note that we can probably expand this technique to work under level 2 if it comes up, but I don't think it's necessary until/unless someone actually complains about it.

Come to think of it, could maybe even do something with purely LD_PRELOAD... which i guess would work under level 3 too. Well, we can leave that for another time.

DarkRTA:
closing this as we are going to handle this in a much better way

PR #378, opened 2020-12-27
Switch Text Shaping to rustybuzz

CryZe:
rustybuzz is a more or less complete port of harfbuzz (state of the art) text shaping engine to Rust. This means our font handling is now about as good as the one found in a browser, except all the manual things you may need to do on top such as using different fonts for different parts of the text, because some glyphs are not available in the chosen font and handling image based glyphs such as those that are being used for emojis in some fonts. Overall though switching to rustybuzz seems to have already changed the appearance quite a bit. There is now proper kerning (though I thought rusttype already did that, but apparently not), proper ligature support and because this is state of the art, it also supports tabular numbers for example which is now activated for all the numbers that we are rendering. Unfortunately since rustybuzz does so much more work, it also heavily slowed down text shaping, which we do on every frame at the moment. So we really need to refactor the renderer to start caching a lot more stuff.


tests/rendering.rs
```diff
@@ -241,6 +241,7 @@ fn check(state: &LayoutState, expected_hash_data: &str, name: &str) {
     check_dims(state, [300, 500], expected_hash_data, name);
 }
 
+#[track_caller]
 fn check_dims(state: &LayoutState, dims: [usize; 2], expected_hash_data: &str, name: &str) {
```
CryZe, line 245: This is a cool change. If the check fails, the error is marked where the caller called this function.

CryZe:
Oops tabular numbers on the timer itself reveals some ugly digits. Time to update the timer font.

wooferzfg:
The Dummy Rendering benchmarks are running into a lot of the following warnings in GitHub Actions:
```
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.6s, enable flat sampling, or reduce sample count to 60.
```
Not sure if this is a problem?

CryZe:
The benchmarks there barely matter, the idea was to upload them to some site, but the site turned out to not work well at all. Also I don't think this is that much of a problem, just means we really should focus on caching this stuff.

src/rendering/font.rs
```diff
@@ -1,195 +1,303 @@
...
-            g.position().x + g.unpositioned().h_metrics().advance_width - first.unwrap()
+            *glyph = the_glyph as _;
+
+            let width = self.font.glyph_x_advance(the_glyph as _);
+            if width > digit_width {
+                digit_width = width;
+            }
+        }
+
+        // FIXME: There's kerning between e.g. ".1" now, which is maybe not quite
```
wooferzfg, line 238: Do you have an example of what this looks like?

CryZe: This is before I switched to tabular nums, but you can kind of see it here how .2 and .1 are differently aligned:

CryZe: If the font (which you can't change yet) doesn't have tnum support, then this will still happen.

CryZe: Kerning could be turned off, but we currently have tnum active for all the values of the key value components, even if they are "info text" components. So turning kerning off for components like Current Comparison would be suboptimal. Though we could introduce a flag in the key value component state that marks it as a "info time" as opposed to a "info text" component.

PR #382, opened 2020-12-31
Add ARM macOS target to CI

CryZe:
Rust 1.49 just released with support for the new macs that use ARM CPUs. Let's add this to our CI.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -468,7 +468,7 @@ jobs:
 
           # macOS
           - label: macOS aarch64
-            toolchain: stable-aarch64-apple-darwin
+            target: aarch64-apple-darwin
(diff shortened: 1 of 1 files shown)


CryZe:
Alright, that's a green build. Unfortunately it seems like GitHub Actions neither has any ARM nodes nor does it have very many BigSur / macOS 11 nodes, meaning that it takes a lot of time for those nodes to boot up. Not sure if we should maybe wait a bit before we merge this, as it just doesn't seem like it's super ready on GitHub's side.

CryZe:
They are tracking performance here: https://github.com/actions/virtual-environments/issues/2486
And native ARM support here: https://github.com/actions/virtual-environments/issues/2187

PR #385, opened 2021-01-06
Implement Support for Custom Fonts

### LINKED ISSUE #104: Customizable Fonts
For supporting fonts, we probably want to express them as Font Name + Font Size (and maybe bold, italic, ..., not entirely sure). That should be the most portable representation. I don't think livesplit-core itself needs to be able to provide a list of all the fonts available. I think the UI can do that. For the browser we probably want to just provide a list of common fonts and allow providing a custom font name, as there doesn't seem to be any API (other than a hack through flash) to actually query all of the system's fonts. In the desktop versions we can use some native APIs to query them. So actually, maybe it does make sense to do that in livesplit-core, in a livesplit-font library or something.

CryZe:
This implements the support for specifying custom fonts in layouts. Unlike the original LiveSplit, this acts more as an override here. One reason for this is to make it easier for platforms to just not have any font support in the first place, such as terminal applications where it's simply not possible to change fonts. If a user never specifies a custom font, it's expected that the application can then use any font that is appropriate for the situation.

Fonts are specified as a font family name and additional styling information, such as the style (italic / regular), weight and stretch of the specific font to select out of the font family. The application then chooses the most appropriate font that it can find, falling back to its default fonts if it can't find the font. It turns out that different APIs and applications have very different notions of what a family name even is. This is because a font can have lots of different, even conflicting, names in its name table. The kinds of names that you specify there have evolved over the years, so there's old entries and newer entries, where the font selector has to make very unclear decisions which to prefer. It seems like modern APIs prefer to select for the `Typographic Family Name` (Name ID 16), with a fallback to `Font Family Name` (Name ID 1) in case the former is not available. Old APIs such as GDI do their own special thing though, complicating the matter even more. I have yet to fully grasp what GDI even does. Even among browsers the font selection seems to work differently. So it all seems like a big mess where you can only try to align with as many APIs as closely as possible, and a correct solution is simply not possible. Since the original LiveSplit seems to select its fonts through GDI, we also run into the issue of importing these identifiers properly. Technically we could query GDI for the specific font, but this is only possible on Windows and is not that easy, as we'd need to parse out the typographic family name, but GDI doesn't have an API for that, nor does it give you the path of the font or access to the font's data. It gives you access to the raw data of individual tables you query for, but ttf-parser doesn't expose a way to parse individual tables in its public
API. So for all of these reason, for now we only implement a fallback converter for these identifiers where we attempt to pop individual style tokens off the end of the identifier to turn it into styling information and the family name. This is however also flawed in other ways, such as those GDI font identifiers also being limited to 32 characters, meaning that you can end up unlucky and have a style token be cut off in the middle at the end, at which point we can't recognize it as such anymore.

Another issue was vertically aligning text. Fonts have various vertical lines, such as the ascender (being the very top a glyph can reach), the baseline (the line that glyphs are written on top of), and the descender (the very bottom a glyph can reach). These three lines can be arbitrarily defined by a font. It is impossible to take two fonts and apply a linear transformation to it to align all three of these lines. Considering we usually have the text font and the times font in the same line, we however still need to make them look aligned. Also there is the possibility that more fonts get involved as certain glyphs may not be available in these fonts, such as emojis. At this point you definitely want to align their baselines. If you also keep a fixed scaling across the fonts, which is usually what you also want, you end up with ascenders and descenders that are all over the place. However this is basically a necessity, there's nothing you can do about this without harming the previously mentioned important properties. Therefore it can be concluded that any sort of attempt at vertical centering is basically impossible and you instead need to choose a baseline that looks decent across many different fonts. Coincidentally this is already what the renderer did, so nothing needed to be changed in this regard. Though we may want to take a closer look at possibly better positioning for that baseline.

The rendering unit tests aren't allowed to use any custom fonts, as that otherwise causes those tests to fail on different platforms where these fonts may not be available. Also this introduces a new `font-loading` feature where our renderer only loads any fonts if the feature is active. This is because the renderer could be compiled for a platform where the code isn't able to load any fonts for, such as in my experiments to compile the renderer to the GameCube and Gameboy Advance. So this is another reason why we simply can't load any custom fonts in our tests.

Resolves #104


src/rendering/component/title.rs
```diff
@@ -49,32 +49,32 @@ pub(in crate::rendering) fn render<B: Backend>(
...
     } else {
-        (height / 2.0 + TEXT_ALIGN_CENTER, line2_end_x - PADDING)
+        (height / 2.0 + TEXT_ALIGN_CENTER, line2_end_x)
     };
 
     let line1 = context.choose_abbreviation(
         component.line1.iter().map(|a| &**a),
         DEFAULT_TEXT_SIZE,
-        width - PADDING - left_bound,
+        line1_end_x - left_bound,
```
CryZe, line 77: For choosing the abbreviation here, we accidentally took the right border of the title component instead of taking into account that there might only be a single line and us sharing the line with the attempt count then. In that case the abbreviation needs to be chosen in respect to how much space it takes up (which is `line1_end_x`).

src/layout/parser.rs
```diff
@@ -103,10 +117,13 @@ impl GradientType for GradientKind {
...
+        // FIXME: Implement delta color support properly:
+        // https://github.com/LiveSplit/livesplit-core/issues/380
+
         Ok(match kind {
-            b"Plain" => GradientKind::Plain,
-            b"Vertical" => GradientKind::Vertical,
-            b"Horizontal" => GradientKind::Horizontal,
+            b"Plain" | b"PlainWithDeltaColor" => GradientKind::Plain,
+            b"Vertical" | b"VerticalWithDeltaColor" => GradientKind::Vertical,
+            b"Horizontal" | b"HorizontalWithDeltaColor" => GradientKind::Horizontal,
```
CryZe, line 126: Turns out that I completely forgot that this was a thing. We completely failed parsing those layouts. For now it kind of just ignores it. There's a proper issue tracking support for them.

CryZe:
Oh god, font-kit really doesn't like to be cross compiled... ooof

src/layout/general_settings.rs
```diff
@@ -9,6 +11,15 @@ use serde::{Deserialize, Serialize};
 pub struct GeneralSettings {
     /// The direction which the components are laid out in.
     pub direction: LayoutDirection,
+    /// The font to use for the timer text. `None` means a default font should
+    /// be used.
+    pub timer_font: Option<Font>,
+    /// The font to use for the times and other values. `None` means a default
+    /// font should be used.
```
wooferzfg, line 18: I'm not entirely sure about this:
```suggestion
    /// The font to use for the times and other numerical values. `None` means a
    /// default font should be used.
```

src/settings/font.rs
```diff
@@ -0,0 +1,160 @@
+use crate::platform::prelude::*;
+use serde::{Deserialize, Serialize};
+
+/// Describes a Font to visualize text with. Depending on the platform a font
+/// that matches the settings most closely is chosen. The settings may be
+/// ignored entirely if the platform can't support different fonts such as in a
+/// terminal for example.
```
wooferzfg, line 7:
```suggestion
/// Describes a Font to visualize text with. Depending on the platform, a font
/// that matches the settings most closely is chosen. The settings may be
/// ignored entirely if the platform can't support different fonts, such as in a
/// terminal.
```

src/settings/font.rs
```diff
@@ -0,0 +1,160 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct Font {
+    /// The family name of the font to use. This corresponds with the
+    /// `Typographic Family Name` (Name ID 16) in the name table of the font. If
+    /// no such entry exists, the `Font Family Name` (Name ID 1) is to be used
+    /// instead. If there are multiple entries for the name, the english entry
+    /// is the one to choose. The subfamily is not specified at all and instead
```
wooferzfg, line 14:
```suggestion
    /// is the one to choose. The subfamily is not specified at all, and instead
```

src/settings/font.rs
```diff
@@ -0,0 +1,160 @@
...
+            Weight::SemiBold => 600.0,
+            Weight::Bold => 700.0,
+            Weight::ExtraBold => 800.0,
+            Weight::Black => 900.0,
+            Weight::ExtraBlack => 950.0,
+        }
+    }
+}
+
+/// The stretch specifies how wide a font should be. For example it may make
```
wooferzfg, line 103:
```suggestion
/// The stretch specifies how wide a font should be. For example, it may make
```

src/rendering/font.rs
```diff
@@ -202,9 +285,24 @@ impl<'f> Glyphs<'f> {
...
-        cursor.x -= adv_x as f32 * self.font.scale / 2.0;
+        let width = adv_x as f32 * self.font.scale;
+
+        // Since we want to delegate to left aligned, we calculate the left
+        // coordinates.
+        cursor.x -= width / 2.0;
         cursor.y -= adv_y as f32 * self.font.scale / 2.0;
 
+        // However we may overlap on the right, in that case, we want to align
+        // to the right instead.
```
wooferzfg, line 296:
```suggestion
        // However. we may overlap on the right. In that case, we want to align
        // to the right instead.
```

src/layout/parser.rs
```diff
@@ -242,6 +259,171 @@ where
...
+            .map_err(|_| Error::ParseFont)?
+            .trim();
+
+        let mut style = FontStyle::Normal;
+        let mut weight = FontWeight::Normal;
+        let mut stretch = FontStretch::Normal;
+
+        // The original LiveSplit is based on Windows Forms, which is just a
+        // .NET wrapper around GDI+. It's a pretty old API from before
+        // DirectWrite existed and fonts used to be very different back then.
```
wooferzfg, line 316:
```suggestion
        // DirectWrite existed, and fonts used to be very different back then.
```

src/layout/parser.rs
```diff
@@ -242,6 +259,171 @@ where
...
+        // This is why GDI uses a very different identifier for fonts than
+        // modern APIs. Since all the modern APIs take a font family, we somehow
+        // need to convert the font identifier from the original LiveSplit into
+        // a font family. The problem is that we may not necessarily even have
+        // the font, nor be on a platform where we could even query for any
+        // fonts or get enough metadata about them, such as in the browser. So
+        // for those case we implement a very simple, though also really lossy
```
wooferzfg, line 323:
```suggestion
        // for those cases, we implement a very simple, though also really lossy
```

src/layout/layout_state.rs
```diff
@@ -10,6 +12,15 @@ pub struct LayoutState {
     pub components: Vec<ComponentState>,
     /// The direction which the components are laid out in.
     pub direction: LayoutDirection,
+    /// The font to use for the timer text. `None` means a default font should
+    /// be used.
+    pub timer_font: Option<Font>,
+    /// The font to use for the times and other values. `None` means a default
+    /// font should be used.
```
wooferzfg, line 19: Same thing here, not entirely sure:
```suggestion
    /// The font to use for the times and other numerical values. `None` means a
    /// default font should be used.
```

src/layout/parser.rs
```diff
@@ -242,6 +259,171 @@ where
...
+                stretch = FontStretch::SemiExpanded;
+            } else if token.eq_ignore_ascii_case("expanded") {
+                stretch = FontStretch::Expanded;
+            } else if token.eq_ignore_ascii_case("extraexpanded") {
+                stretch = FontStretch::ExtraExpanded;
+            } else if token.eq_ignore_ascii_case("ultraexpanded") {
+                stretch = FontStretch::UltraExpanded;
+            } else if !token.eq_ignore_ascii_case("regular") {
+                family =
+                    &family[..token.as_ptr() as usize - family.as_ptr() as usize + token.len()];
```
wooferzfg, line 397: What does this line do?

src/layout/general_settings.rs
CryZe: It's use for all values of key value components like current comparison. At least at the moment.

src/layout/parser.rs
CryZe: So this is where we notice that the token we encountered is part of the family name and we should stop scanning. So what we want here is to assign the family name up until and including the token we just looked at. Sometimes iterators in Rust have the ability to give you access to the remaining underlying data, but somehow not .split_whitespace(), but even then it wouldn't include the current item anyway, only the rest. So what I'm doing instead is some pointer shenanigans (still all completely safe). Basically I take the token's pointer and turn it into an integer address and subtract the address of the whole family name. That is how many bytes into the whole string the token is. If we add the length of the token to it (since we want to include it, as it turned out to not be a style token) we get the total length from the beginning of the whole string that we consider the family name. We then take a substring based on that length.

src/rendering/font.rs
CryZe: However. ?

wooferzfg: Sorry, should be `However,`

src/layout/general_settings.rs
wooferzfg: Alright, we can keep it as is then.

PR #392, opened 2021-01-18
Always add content after parentheses in extended category

### LINKED ISSUE #391: Closing parenthesis doesn't work for category name
If you put `Any% (KPDR27)` as the category name, you get `Any% (KPDR27` in the title component, missing the closing parenthesis.

wooferzfg:
I'm not sure if I'm missing an edge case with this fix, but this should only impact situations where `has_pushed` is `false`, which only happens when no additional information is added to the extended category name.

Fix #391

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ base64 = { version = "0.13.0", default-features = false, features = ["alloc"] }
 cfg-if = "1.0.0"
 chrono = { version = "0.4.10", default-features = false, features = ["serde", "alloc"] }
 derive_more = { version = "0.99.1", default-features = false, features = ["not", "add", "from", "deref"] }
-hashbrown = "0.10.0"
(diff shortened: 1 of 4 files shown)


CryZe:
I'd like to see some tests here.

CryZe:
Uhm what

wooferzfg:
Not sure what's going on here:
`error: failed to select a version for the requirement hashbrown = "^0.10.0"`

CryZe:
They yanked the new major version that they released. It probably had a major bug. But lmao, that's an oversight on cargo's implementation imo. Yanking a whole major version range shouldn't be possible.

CryZe:
@wooferzfg Can you basically revert #394 as part of this?

CryZe:
Seems like this is the cause: https://github.com/rust-lang/hashbrown/issues/225

CryZe (approved):

PR #395, opened 2021-01-21
C API: Expose the Layout Editor's State.

DarkRTA:
This really doesn't need too much explaining, but do note I went the route of exposing `SettingValue` as JSON to avoid having to create 20 different getter functions.

### DIFF
--- a/capi/src/layout_editor.rs
+++ b/capi/src/layout_editor.rs
@@ -6,6 +6,7 @@
 use super::{output_vec, Json};
 use crate::component::OwnedComponent;
 use crate::layout::OwnedLayout;
+use crate::layout_editor_state::OwnedLayoutEditorState;
 use crate::setting_value::OwnedSettingValue;
 use livesplit_core::{LayoutEditor, Timer};
 
@@ -36,6 +37,12 @@ pub extern "C" fn LayoutEditor_state_as_json(this: &LayoutEditor) -> Json {
     output_vec(|o| {
         this.state().write_json(o).unwrap();
     })
+}
+
+/// Returns the state of the Layout Editor.
+#[no_mangle]
+pub extern "C" fn LayoutEditor_state(this: &LayoutEditor) -> OwnedLayoutEditorState {
+    Box::new(this.state())
 }
 
 /// Encodes the layout's state as JSON based on the timer provided. You can use
--- a/capi/src/layout_editor_state.rs
+++ b/capi/src/layout_editor_state.rs
@@ -1,0 +1,96 @@
+//! Represents the current state of the Layout Editor in order to visualize it properly.
+use crate::{c_char, output_str};
+use livesplit_core::{layout::editor::State as LayoutEditorState, settings::Value as SettingValue};
+
+/// type
+pub type OwnedLayoutEditorState = Box<LayoutEditorState>;
+/// type
+pub type NullableOwnedLayoutEditorState = Option<OwnedLayoutEditorState>;
+
+/// drop
+#[no_mangle]
+pub extern "C" fn LayoutEditorState_drop(this: OwnedLayoutEditorState) {
+    drop(this);
+}
+
+/// Returns the number of components in the layout.
+#[no_mangle]
+pub extern "C" fn LayoutEditorState_component_len(this: &LayoutEditorState) -> usize {
+    this.components.len()
+}
+
+/// Returns the name of the component at the specified index.
+#[no_mangle]
+pub extern "C" fn LayoutEditorState_component_text(
+    this: &LayoutEditorState,
+    index: usize,
+) -> *const c_char {
+    output_str(&this.components[index])
+}
+
+/// Returns a bitfield corresponding to which buttons are active.
+///
+/// The bits are as follows:
+///
+/// * `0x04` - Can remove the current component
+/// * `0x02` - Can move the current component up
+/// * `0x01` - Can move the current component down
+#[no_mangle]
+pub extern "C" fn LayoutEditorState_buttons(this: &LayoutEditorState) -> u8 {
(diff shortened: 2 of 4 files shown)


capi/src/layout_editor_state.rs
```diff
@@ -0,0 +1,96 @@
...
+///
+/// * `0x04` - Can remove the current component
+/// * `0x02` - Can move the current component up
+/// * `0x01` - Can move the current component down
+#[no_mangle]
+pub extern "C" fn LayoutEditorState_buttons(this: &LayoutEditorState) -> u8 {
+    this.buttons.can_remove as u8 * 0x4
+        | this.buttons.can_move_up as u8 * 0x2
+        | this.buttons.can_move_down as u8 * 0x1
+}
```
DarkRTA, line 43: There is almost certainly a better way of doing this besides packing everything into a `u8`.

src/settings/value.rs
```diff
@@ -72,6 +72,22 @@ pub enum Error {
...
+    }
+
+    /// Encodes the value as JSON.
+    pub fn write_json<W>(&self, writer: W) -> serde_json::Result<()>
+    where
+        W: std::io::Write,
+    {
+        serde_json::to_writer(writer, self)
+    }
+
```
DarkRTA, line 90: This is a bit unrelated but I feel like a derive macro to automatically add these would be extremely useful for various places in the codebase.

capi/src/layout_editor_state.rs
CryZe: Seems fine to me, though I'd usually shift << 1, << 2 and co. rather than multiply.

capi/src/layout_editor_state.rs
```diff
@@ -0,0 +1,96 @@
...
+/// Returns the name of the component at the specified index.
+#[no_mangle]
+pub extern "C" fn LayoutEditorState_components_text(
+    this: &LayoutEditorState,
+    index: usize,
+) -> *const c_char {
+    output_str(&this.components[index])
+}
+
+/// Outputs a bitfield corresponding to which buttons are active.
```
CryZe, line 31: We usually say "Returns" rather than "Outputs"

capi/src/layout_editor.rs
```diff
@@ -4,6 +4,7 @@
 //! state objects that can be visualized by any kind of User Interface.
 
 use super::{output_vec, Json};
+use crate::layout_editor_state::OwnedLayoutEditorState;
```
CryZe, line 7: Seems like cargo fmt wants this reordered, doesn't matter too much though.

capi/src/setting_value.rs
```diff
@@ -1,7 +1,7 @@
 //! Describes a setting's value. Such a value can be of a variety of different
 //! types.
 
-use crate::str;
+use crate::{str, Json, output_vec};
```
CryZe, line 4: Cargo fmt unhappy again.

capi/src/layout_editor_state.rs
```diff
@@ -0,0 +1,96 @@
...
+
+/// Returns the number of components in the layout.
+#[no_mangle]
+pub extern "C" fn LayoutEditorState_components_len(this: &LayoutEditorState) -> usize {
+    this.components.len()
+}
+
+/// Returns the name of the component at the specified index.
+#[no_mangle]
+pub extern "C" fn LayoutEditorState_components_text(
```
CryZe, line 24: I think I rather prefer the singular `component_text` here (as well for the others), as you query for a specific component.

capi/src/layout_editor_state.rs
```diff
@@ -0,0 +1,96 @@
...
+/// * `0x02` - Can move the current component up
+/// * `0x01` - Can move the current component down
+#[no_mangle]
+pub extern "C" fn LayoutEditorState_buttons(this: &LayoutEditorState) -> u8 {
+    this.buttons.can_remove as u8 * 0x4
+        | this.buttons.can_move_up as u8 * 0x2
+        | this.buttons.can_move_down as u8 * 0x1
+}
+
+/// The index of the currently selected component
```
CryZe, line 45: This one is the only one that isn't a full sentence (lack of . at the end) and doesn't have a verb ("Returns ...").

capi/src/layout_editor_state.rs
```diff
@@ -0,0 +1,96 @@
...
+#[no_mangle]
+pub extern "C" fn LayoutEditorState_selected_component(this: &LayoutEditorState) -> u32 {
+    this.selected_component
+}
+
+/// Gets the number of fields in the layout's settings.
```
CryZe, line 51: Not sure if we use "Gets ..." anywhere else.

CryZe (changes requested):
It's honestly mostly fine. Just some very minor things. You honestly don't even need to fix all of them if you don't want. Just felt weird to only point out one minor thing and not also point out the others.

capi/src/layout_editor_state.rs
DarkRTA: I'll probably leave this as multiplication so it matches the documentation. Both should optimize to the same code anyway.

capi/src/layout_editor.rs
DarkRTA: I meant to run `cargo fmt` before opening the PR. Whoops.

capi/src/layout_editor_state.rs
DarkRTA: I'll adjust the naming of all of the functions meant to query a `Vec<_>` to use singular names.

DarkRTA:
I'm pretty sure I've addressed everything here. Feel free to merge when CI finishes.

PR #422, opened 2021-04-10
Implement Hotkeys for macOS

### LINKED ISSUE #24: Hotkey Support for other Systems
- [x] Windows
- [x] Linux
- [ ] macOS
- [x] emscripten

CryZe:
This implements the global hotkeys for macOS. The code is mostly based on:

https://github.com/kwhat/libuiohook/blob/f4bb19be8aee7d7ee5ead89b5a89dbf440e2a71a/src/darwin/input_hook.c#L1086

There's some stuff I left out, but it doesn't seem necessary.

Additionally I renamed all of the `NumPad*` variants of the individual `KeyCode` enums to `Numpad*`. We probably really only want a single enum that is used by all the platforms, so there's no such disparities.

Resolves #24


wooferzfg:
Has this been tested on a Mac already?

CryZe:
It works in my VM at least

crates/livesplit-hotkey/src/linux/key_code.rs
```diff
@@ -658,41 +658,41 @@ impl FromStr for KeyCode {
 //     Break = 0xFF6B,
 //     ModeSwitch = 0xFF7E,
 //     NumLock = 0xFF7F,
-//     NumPadSpace = 0xFF80,
```
wooferzfg, line 661: What's the reason for maintaining this big list?

crates/livesplit-hotkey/src/macos/key_code.rs
```diff
@@ -0,0 +1,262 @@
...
+            "F13" => F13,
+            "F14" => F14,
+            "F15" => F15,
+            "F16" => F16,
+            "F17" => F17,
+            "F18" => F18,
+            "F19" => F19,
+            "F20" => F20,
+
+            "Numpad0" => Numpad0,
```
wooferzfg, line 194: Is the goal to have these strings be consistent across operating systems?

wooferzfg (approved):

crates/livesplit-hotkey/src/linux/key_code.rs
CryZe: I actually have no idea what this is even for either. I'll clean up the KeyCode enums (and merge them into one) in a subsequent PR.

crates/livesplit-hotkey/src/macos/key_code.rs
CryZe: Yeah, otherwise using the library (like us setting NumPad* to be the default for splitting / resetting / ...) is just unnecessarily painful.

PR #429, opened 2021-05-26
Add Current Split Index to the C API

MichaelJBerk:


capi/src/timer.rs
```diff
@@ -77,6 +77,20 @@ pub extern "C" fn Timer_set_run(this: &mut Timer, run: OwnedRun) -> NullableOwne
     this.set_run(*run).err().map(Box::new)
 }
 
+/// Accesses the index of the split the attempt is currently on. If there's
+/// no attempt in progress, `None` is returned instead. This returns an
```
CryZe, line 81:
```suggestion
/// no attempt in progress, `-1` is returned instead. This returns an
```

capi/src/timer.rs
```diff
@@ -77,6 +77,20 @@ pub extern "C" fn Timer_set_run(this: &mut Timer, run: OwnedRun) -> NullableOwne
...
+/// index that is equal to the amount of segments when the attempt is
+/// finished, but has not been reset. So you need to be careful when using
+/// this value for indexing.
+#[no_mangle]
+pub extern "C" fn Timer_current_split_index(this: &Timer) -> i64 {
```
CryZe, line 86: I'd probably prefer isize as the return type, simply because i64 is a semi weird type on WASM and anything length / index should usually be usize / isize.

CryZe:
Solaris doesn't work anymore. Feel free to comment it out. If not I can do it too.

CryZe:
It's fixed on master now.

PR #430, opened 2021-06-08
Introduce Layers in the Renderer

CryZe:
There are a lot of elements that are basically almost entirely static on the layout. Other elements such as the timer are frequently changing. In the browser we already noticed that by pushing the timer onto its own layer, you get a fairly big performance boost out of it.

Here's that Pull Request:

https://github.com/LiveSplit/LiveSplitOne/pull/378

We want to apply a similar idea to the native renderer. In order to do this we first introduce flags in the layout state that mark whether a component or a part of a component is at the specific time considered frequently changing. By doing this not only does the native renderer benefit from it, but other renderers such as the web version can also make decisions based on that, as opposed to only applying the optimization to the timer component.

With all the components providing these flags, the renderer can now split up the frame into a bottom layer and a top layer where all the frequently changing elements are rendered onto the top layer and all the other elements rendered onto the bottom layer. In fact the top layer is actually the frame buffer and it simply gets cleared by copying over regions from the bottom layer. That way there is no additional compositing necessary.

Additionally the renderer is now more of a scene manager that manages entities in a scene rather than directly emitting draw calls to a backend. The scene is a data structure that allows traversing the two layers in any way a renderer (i.e. what used to be a backend) would want to render them out. So the whole design is a lot more decoupled now.


src/analysis/current_pace.rs
```diff
@@ -7,7 +7,7 @@ use crate::{analysis, timing::Snapshot, TimeSpan, TimerPhase};
 /// Calculates the current pace of the active attempt based on the comparison
 /// provided. If there's no active attempt, the final time of the comparison is
 /// returned instead.
```
wooferzfg, line 9: Should we also have something like this here? "Additionally a boolean is returned that indicates if the value is currently actively changing as time is being lost."

src/rendering/software.rs
```diff
@@ -173,143 +79,413 @@ impl Backend for SoftwareBackend<'_> {
...
+        let max_y = mem::replace(&mut self.max_y, max_y).max(max_y);
+
+        if force_redraw || bottom_layer_changed {
+            frame_buffer
+                .data_mut()
+                .copy_from_slice(background.data_mut());
+        } else if min_y <= max_y {
+            let stride = 4 * stride as usize;
+            let min_y = stride * (min_y - 1.0) as usize;
+            let max_y = stride * ((max_y + 2.0) as usize).min(height as usize);
```
wooferzfg, line 211: What's the reasoning behind the calculations here? I guess I don't really understand where these numbers are coming from.

src/analysis/current_pace.rs
CryZe: Yeah that should've probably mentioned it. I'll try to sneak it into some future commit.

Issue #424, opened 2021-04-12
Remove `derive_more`


CryZe:
It's definitely convenient, but proc macros are usually fairly expensive dependencies, so we probably should just write the few trait impls ourselves.

PR #433, opened 2021-06-13
remove derive_more

ash2x3zb9cy:
Issue #424 suggests removing derive_more and writing out the impls explicitly, which is what this commit does.


CryZe:
Thanks a lot, this is great. :)

PR #436, opened 2021-06-29
Add support for Property layoutPath in Run-Class

Relacibo:
Needed for LiveSplit/LiveSplit#2145

### DIFF
--- a/src/run/mod.rs
+++ b/src/run/mod.rs
@@ -81,6 +81,7 @@ pub struct Run {
     custom_comparisons: Vec<String>,
     comparison_generators: ComparisonGenerators,
     auto_splitter_settings: Vec<u8>,
+    layout_path: String,
 }
 
 #[derive(Clone, Debug)]
@@ -125,6 +126,7 @@ impl Run {
             custom_comparisons: vec![personal_best::NAME.to_string()],
             comparison_generators: ComparisonGenerators(default_generators()),
             auto_splitter_settings: Vec::new(),
+            layout_path: String::new(),
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/run/mod.rs
```diff
@@ -336,6 +338,22 @@ impl Run {
         &mut self.auto_splitter_settings
     }
 
+    /// Layout, the run is using.
```
wooferzfg, line 341:
```suggestion
    /// Path to the layout associated with the run.
```

src/run/mod.rs
```diff
@@ -336,6 +338,22 @@ impl Run {
...
+    #[inline]
+    pub fn layout_path(&self) -> &str {
+        &self.layout_path
+    }
+
+    /// Sets the layout, the run is using.
```
wooferzfg, line 347:
```suggestion
    /// Sets the path to the layout associated with the run.
```

wooferzfg (commented):
We should update the run saver as well: https://github.com/LiveSplit/livesplit-core/blob/ec5486d76378d25e26226196c1e4af6b202a2432/src/run/saver/livesplit.rs

Relacibo:
89e5125 is the commit that works. After that i experimented with enum for filepath(path) default, but the capi doesn't it seems like doesn't support enums and i stopped there pretty much.

DarkRTA:
You are correct in the C API lacking support for  `enum`. When I needed an `enum`, my options were to expose it as a string or expose it as an int while documenting the possible values.

CryZe:
I'll look into this for the Desktop version of LiveSplit One.

CryZe:
What happened?!

CryZe:
I believe the PR "corrupted" because you used the master branch, so me rebasing the PR caused it to be detected as being the same as the master branch now and therefore it auto closed it?! That's so weirdly broken.

PR #449, opened 2021-10-01
Add layout settings for time formatters

doesthisusername:
This PR exposes settings for time accuracy and delta time's "drop decimals past one minute" thing.

I don't know if having the settings layout-global is ideal (probably not), but I wasn't sure what else to pass in column's `update_state` if it were local to each `Splits` component.
I also didn't change any parsing code for the original LiveSplit layouts, as I'm not sure where my changes would even be...

Tests seem to pass too.

(feel free to give suggestions if i should change anything)

### DIFF
--- a/src/component/splits/column.rs
+++ b/src/component/splits/column.rs
@@ -2,13 +2,13 @@ use crate::{
     analysis::{self, possible_time_save, split_color},
     clear_vec::Clear,
     comparison,
+    component::splits::Settings as SplitsSettings,
     platform::prelude::*,
     settings::{Color, SemanticColor},
     timing::{
         formatter::{Delta, Regular, SegmentTime, TimeFormatter},
         Snapshot,
     },
-    component::splits::Settings as SplitsSettings,
     GeneralLayoutSettings, Segment, TimeSpan, TimingMethod,
 };
 use core::fmt::Write;
@@ -144,21 +144,21 @@ enum ColumnFormatter {
 
 pub fn update_state(
     state: &mut ColumnState,
-    column: &ColumnSettings,
+    column_settings: &ColumnSettings,
     timer: &Snapshot<'_>,
-    splits: &SplitsSettings,
+    splits_settings: &SplitsSettings,
     layout_settings: &GeneralLayoutSettings,
     segment: &Segment,
     segment_index: usize,
     current_split: Option<usize>,
     method: TimingMethod,
 ) {
-    let method = column.timing_method.unwrap_or(method);
-    let resolved_comparison = comparison::resolve(&column.comparison_override, timer);
+    let method = column_settings.timing_method.unwrap_or(method);
+    let resolved_comparison = comparison::resolve(&column_settings.comparison_override, timer);
     let comparison = comparison::or_current(resolved_comparison, timer);
 
     let update_value = column_update_value(
-        column,
+        column_settings,
         timer,
         segment,
         segment_index,
@@ -171,7 +171,7 @@ pub fn update_state(
 
     let ((column_value, semantic_color, formatter), is_live) = update_value.unwrap_or_else(|| {
         (
-            match column.start_with {
+            match column_settings.start_with {
                 ColumnStartWith::Empty => (None, SemanticColor::Default, ColumnFormatter::Time),
                 ColumnStartWith::ComparisonTime => (
                     segment.comparison(comparison)[method],
@@ -198,7 +198,7 @@ pub fn update_state(
         )
     });
 
-    let is_empty = column.start_with == ColumnStartWith::Empty && !updated;
+    let is_empty = column_settings.start_with == ColumnStartWith::Empty && !updated;
 
     state.updates_frequently = is_live && column_value.is_some();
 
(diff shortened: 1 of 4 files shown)


src/layout/general_settings.rs
```diff
@@ -48,6 +49,15 @@ pub struct GeneralSettings {
...
     pub text_color: Color,
+    /// Specifies the display accuracy of split times.
+    pub split_time_accuracy: Accuracy,
+    /// Specifies the display accuracy of segment times.
+    pub segment_time_accuracy: Accuracy,
+    /// Specifies the display accuracy of delta times.
+    pub delta_time_accuracy: Accuracy,
+    /// Whether to drop the fractional part of a delta time once it goes past
+    /// one minute.
+    pub delta_drop_decimals: bool,
```
CryZe, line 60: Yeah these should be moved to the settings of the Splits component: https://github.com/LiveSplit/livesplit-core/blob/master/src/component/splits/mod.rs#L48

You can pass these settings to the columns function as a parameter.

These settings then also need to be exposed in the layout editor via these two functions: https://github.com/LiveSplit/livesplit-core/blob/master/src/component/splits/mod.rs#L434-L552

CryZe (changes requested):
Additionally you want to parse it properly from the original LiveSplit's layouts: https://github.com/LiveSplit/livesplit-core/blob/master/src/layout/parser/splits.rs#L178-L179

src/layout/general_settings.rs
doesthisusername: Oh right, thank you! I'll try to get this done tomorrow. Would you recommend passing the whole of the split settings to the column update, or just those relevant parts? dunno how tightly coupled columns and splits are allowed/supposed to be

CryZe: Passing the whole thing would keep the argument list smaller (and thus faster), so I'd go with that.

doesthisusername:
I think that should be it?
Though, I'm personally not a big fan of replicating the defaults of the different formatters in the default impl of the splits settings now (instead of using them directly). Any suggestions/thoughts there?
Another thing is that I can imagine that users might also find the meaning of "segment time" confusing in the layout settings, given how it's sort of nichely used within it. I can imagine a lot of people will just choose the same setting for Split Time and Segment Time, so maybe it's not too big of a deal.

wooferzfg:
> Another thing is that I can imagine that users might also find the meaning of "segment time" confusing in the layout settings, given how it's sort of nichely used within it.

I think it should be fine because "Split Time" and "Segment Time" are the wording used when choosing the column "update with" type as well.

doesthisusername:
oh i forgot to format two of the files, suppose i should commit them formatted?

src/layout/parser/splits.rs
```diff
@@ -155,11 +155,16 @@ where
                             });
                         }
                     })
+                } else if tag.name() == b"SplitTimesAccuracy" {
+                    accuracy(reader, tag.into_buf(), |v| settings.split_time_accuracy = v)
```
CryZe, line 159: Also set it for the segment times here, as there is no such distinction in the original LS, so you would use this value there as well to specify the segment times accuracy.

src/layout/parser/splits.rs
```diff
@@ -1,5 +1,5 @@
 use super::{
-    comparison_override, end_tag, parse_bool, parse_children, text, text_err, text_parsed,
+    accuracy, comparison_override, end_tag, parse_bool, parse_children, text, text_err, text_parsed,
```
CryZe, line 2: Yeah please run `cargo fmt`.

src/component/splits/mod.rs
```diff
@@ -466,6 +480,22 @@ impl Component {
...
+            Field::new(
+                "Segment Time Accuracy".into(),
+                self.settings.segment_time_accuracy.into(),
+            ),
+            Field::new(
+                "Delta Time Accuracy".into(),
+                self.settings.delta_time_accuracy.into(),
+            ),
+            Field::new(
+                "Drop Decimals When More Than 1 Minute".into(),
```
CryZe, line 496: The original LS has these all in groups, so it's obvious what this refers to. Here we don't have any groups yet, so maybe somehow include that this refers to the delta.

src/component/splits/column.rs
```diff
@@ -145,6 +146,7 @@ pub fn update_state(
     state: &mut ColumnState,
     column: &ColumnSettings,
     timer: &Snapshot<'_>,
+    splits: &SplitsSettings,
```
CryZe, line 149: I'm not particularly liking that this name doesn't imply that these are the settings.

CryZe (changes requested):
Looks good for the most part :)

src/component/splits/column.rs
doesthisusername: Yeah, I wasn't sure what to name it either, but ended up mimicking the `column: &ColumnSettings`. Is `splits_settings` better?

CryZe: Oh, yeah maybe rename both then.

src/component/splits/mod.rs
doesthisusername: what do you think of `Drop Delta Decimals When Showing Minutes`? or a more direct change to `Drop Delta Decimals When More Than 1 Minute`?

CryZe: The former seems better.

doesthisusername:
I think I fixed the things pointed out. I guess now the parameter name for `column_update_value`'s `&ColumnSettings` is still `column`, which might be worth changing too, for symmetry.
btw if you want me to squash or something i can try that, not a git guru yet

wooferzfg:
@doesthisusername If you want to squash, the easiest way to do that would be `git rebase -i HEAD~3` (or however many commits you have on your PR) and then force pushing with `git push --force origin head`

doesthisusername:
well, i'd like to if it's preferable for you. i'm off to bed now, so i won't do it right now

CryZe:
Nah, you don't need to squash.

PR #463, opened 2021-10-28
Adds cache to github actions builds

### LINKED ISSUE #297: Possibly add some caching to CI
We are currently not caching anything. Doing so should significantly improve the builds on CI. However I'm not sure if the huge amount of data and jobs we are producing is going to cause any problems, so we'll have to just try it and see.

Eein:
This PR adds the following:

- Generates a Cargo.lock file manually
- Adds a cache for the matrixed builds using the matrix name alongside the lockfile
- Adds a cache for clippy lint 
- Adds a cache for benchmark
- adds restore_keys array to retrieve cache from most recent build if one does not exist for current lock
- Adds workflow dispatch, which allows manually triggering the build on branches or just replaying an existing build; can remove this if asked

Time gains can be seen from:

https://github.com/Eein/livesplit-core/actions/runs/1392897119

to

https://github.com/Eein/livesplit-core/actions/runs/1392944530

Edit: allowing a build to complete here:

https://github.com/Eein/livesplit-core/actions/runs/1393008609

Edit 2: Run with more unique cache keys

https://github.com/Eein/livesplit-core/actions/runs/1393099509

Notes:
- There was some issues with mac builds in my branch, but I do not think its related to this PR and likely some licensing/secrets.
- There's some windows builds are very inconsistent in how long they may take and recompile anyways. Keys may need to be more specific to avoid using caches with the same target name that aren't useful to their context
- cargo packages loaded from git must be recompiled

closes #297

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,6 +2,7 @@ name: Build
 
 on:
   pull_request:
+  workflow_dispatch:
   push:
     branches:
       - 'master'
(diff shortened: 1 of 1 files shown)


CryZe:
Mmh, since there's many (indirect) dependencies I don't think the Cargo.lock will last very long, causing the caches to last probably much less than a day.

Eein:
Sounds good! 

If this doesn't provide any immediate value, Id be happy to have this PR closed.

Can always be picked up later if something changes

CryZe:
Alright, closing this for now. Thanks for looking into this anyway.

PR #467, opened 2021-11-03
Cache Labels in Scene Manager

### LINKED ISSUE #432: Add persistent labels in the scene manager
The scene manager currently shapes all the text on every frame. Turns out this is really expensive. It probably makes up at least half of our frame time.

CryZe:
Instead of shaping all the text every frame, the recently introduced caching system allows us to cache the labels and only shape them again when they change. This gives a hefty performance boost of up to 25x for the scene update.

Resolves #432


CryZe:
And just as I merge this, I notice that I forgot to implement invalidating the labels based on font changes 🤦

PR #477, opened 2021-11-11
Auto-splitting Updates

P1n3appl3:
@TheTedder and I have done some major cleanup of the auto-splitting rework that was living on his branch for the past year or so. We want this to live on the upstream repo so that it's more visible and less likely to keep getting out of sync.

The major changes from the current `auto-splitting` branch are as follows:
- The function interface exposed to the wasm autosplitters has been completely redone, see the doc comment in `src/auto_splitting/mod.rs` for details
- The `process` module has been rewritten using some off the shelf crates that support reading process memory in a platform agnostic way.
- The WIP WASI support and associated tests have been removed. We plan to eventually use WASI but for now we just support `wasm32-unknown-unknown` auto-splitters

Since all of the functionality is behind the non-default `auto-splitting` feature merging it into master should probably be fine, but if y'all want to wait until it's "stabilized" then we can merge this into its own branch.


crates/livesplit-auto-splitting/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "livesplit-auto-splitting"
+version = "0.1.0"
+authors = ["Christopher Serr <christopher.serr@gmail.com>"]
+edition = "2021"
+
+[dependencies]
+time = "0.3.3"
+anyhow = "1.0.45"
```
CryZe, line 9: We likely want `snafu` instead. This was just a little more convenient for prototyping.

src/auto_splitting/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+}
+
+type Result<T> = std::result::Result<T, Error>;
+
+/// The autosplitter runtime is a thread running an event loop. It holds a
+/// shared timer that can be updated by the wasm autosplitter.
+///
+/// The only communication possible with the runtime is to load or unload an
+/// autosplitter. For passing arbitrary data such as configuration info from
+/// inside the autosplitter, use `set_variable()`.
```
CryZe, line 71: That's not what `set_variable` is for.

src/auto_splitting/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+}
+
+// This newtype is required because SharedTimer is an Arc, so we can't impl
+// the autosplitter Timer trait directly on it
+struct AST(SharedTimer);
+
+impl AutoSplitTimer for AST {
+    fn state(&self) -> TimerState {
+        // Safety: these are the same enum
+        unsafe { std::mem::transmute(self.0.read().current_phase()) }
```
CryZe, line 190: That's too unsafe for something that would be a trivial match.

src/auto_splitting/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+
+enum Request {
+    LoadScript(PathBuf, Sender<Result<()>>),
+    UnloadScript(Sender<()>),
+    End,
+}
+
+// This newtype is required because SharedTimer is an Arc, so we can't impl
+// the autosplitter Timer trait directly on it
+struct AST(SharedTimer);
```
CryZe, line 185: Is this even true? Usually you can implement traits if at least part of the type (being the Timer) is defined in your own crate. I might be wrong though. But worth checking again.

CryZe (changes requested):

crates/livesplit-auto-splitting/src/process.rs
```diff
@@ -0,0 +1,71 @@
...
+pub type Result<T> = std::result::Result<T, Error>;
+pub type Address = u64;
+
+#[cfg(target_os = "windows")]
+type OurPid = u32;
+#[cfg(not(target_os = "windows"))]
+type OurPid = i32;
+
+pub struct Process {
+    handle: ProcessHandle,
```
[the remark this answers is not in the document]
P1n3appl3: bugfix: rbspy/read-process-memory#16

crates/livesplit-auto-splitting/src/process.rs
```diff
@@ -0,0 +1,67 @@
...
+        let pid = processes
+            .first()
+            .ok_or(Error::ProcessDoesntExist)
+            .map(|&p| p.pid())? as Pid;
+        let handle = pid.try_into().map_err(|_| Error::ProcessOpening)?;
+        Ok(Process { handle, pid })
+    }
+
+    pub fn is_open(&self, sysinfo: &mut System) -> bool {
+        sysinfo.refresh_processes();
```
CryZe, line 42: I just randomly stumbled upon this: https://blog.guillaume-gomez.fr/articles/2021-12-14+sysinfo%3A+version+0.22+and+FreeBSD+support

Apparently you can refresh just specific processes now.

CryZe: I implemented this in the dependency update commit I believe.

CryZe:
The crate probably needs `#![warn(missing_docs)]` and then all those warnings need to be resolved. I think that's mostly it.

KaDiWa4:
wasmtime doesn't support 32-bit Windows, maybe there should be an extra condition in `.github/workflows/test.sh` to not enable `--all-features`

CryZe:
Process::is_open is absurdly slow. I don't think we can just "refresh all processes".

CryZe (approved):

Issue #65, opened 2017-07-29
Ensure proper handling of Nullable Strings in C API


CryZe:
Seems like we don't properly handle this in most of the language bindings. 

Languages to check / fix:
- [x] JavaScript / TypeScript for emscripten
- [ ] JavaScript / TypeScript for Node
- [x] C#
- [ ] Java JNA
- [ ] Java JNI
- [ ] Kotlin JNI
- [ ] Swift
- [ ] Ruby
- [ ] Python

PR #489, opened 2021-12-11
Rewrite swift API `bind_gen`

KaDiWa4:
Previously, the generated bindings didn't actually work, there was even one syntax error in the swift code (and a lot of wrong types).
I don't know what the `link` lines in the module map did, I hope that doesn't break anything.

# Changes

- Cleanup of `main.rs`
- Rust-style doc comments (`///`)
- Nullable strings as outputs (#65)
- Fix types (remove bool/int  conversions, fix timer lock functions)
- Folder structure works with swift package manager (you need to set the linker `-L` path in the config)
- Remove some of the `override init`s
- Function `default` needs backticks
- Generate class docs only once
- Replace `Optional.none` with `nil`, seems to behave the same with equality
- Rename `LiveSplitCoreNative` to `CLiceSplitCore`
- Rename `module.map` to `module.modulemap` because the former is deprecated
- Move `init?` `nil`-check

### DIFF
--- a/capi/bind_gen/src/main.rs
+++ b/capi/bind_gen/src/main.rs
@@ -13,7 +13,8 @@ mod typescript;
 mod wasm;
 mod wasm_bindgen;
 
-use std::fs::{create_dir_all, remove_dir_all, File};
+use std::collections::BTreeMap;
+use std::fs::{self, create_dir_all, remove_dir_all, File};
 use std::io::{BufWriter, Read, Result};
 use std::path::PathBuf;
 use std::rc::Rc;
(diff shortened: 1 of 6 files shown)


CryZe:
Can you try to rebase away that merge commit? I'd prefer not to merge merge commits. Though I guess GitHub has a way to squash and merge anyway.

CryZe:
There's an empty C file, is this intended?

CryZe:
It seems intentional, but I'd like to see a comment in the code about why this is necessary. Other than that, I think it looks good.

KaDiWa4:
It is intentional, I added a comment

PR #490, opened 2021-12-11
Cache digit width for tabular numbers

### LINKED ISSUE #470: Cache glyph sizes for tabular numbers
A lot of our labels are shaped with tabular numbers. For that we manually position the digits based on the widest digit in the font. At the moment we query the font for those 10 digit glyphs every time we want to update a label, even if it is already shaped. We can almost definitely cache it within the font itself when loading it.

KaDiWa4:
Fixes #470.
The max digit width is now calculated when the font is loaded, even if that font is never used for tabular numbers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ capi/bindings
 Cargo.lock
 docs
 target
+.vs*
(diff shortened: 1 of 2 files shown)


CryZe:
This looks great. Thank you :)

CryZe:
Seems to have been a decent performance improvement:
```
Scene Management (Default)
Before:
                        time:   [7.3866 us 7.4272 us 7.4747 us]
After:
                        time:   [4.3817 us 4.4423 us 4.4991 us]

Scene Management (Subsplits Layout)
Before:
                        time:   [26.779 us 26.782 us 26.786 us]
After:
                        time:   [20.934 us 21.212 us 21.506 us]
```

PR #493, opened 2021-12-26
Rewrite Graph component calculations

KaDiWa4:
The goal of this PR is to make `src/component/graph.rs` easier to understand. To do that, I restructured the functions, which makes the diff very hard to read. Hopefully the new code structure can compensate for that. The functions are mostly executed in the order they're defined in. I didn't make any changes that break the public API, the graph looks almost exactly the same. I tested this with livesplit-one-desktop and the browser LSO version.

Some examples of things that are bad about the `master` version of this file:
- The result of `calculate_left_side_coordinates` is never used
- "Graph" is used to refer to both the graph and the entire diagram in comments
- In a lot of calculations it is hard to tell why some of the variables are used or they have unclear names
- `WIDTH` and `HEIGHT` have arbitrary values (180 and 120) and the graph is later transformed into a `0..1` square

### DIFF
--- a/src/component/graph.rs
+++ b/src/component/graph.rs
@@ -1,7 +1,14 @@
 //! Provides the Graph Component and relevant types for using it. The Graph
 //! Component visualizes how far the current attempt has been ahead or behind
-//! the chosen comparison throughout the whole attempt. All the individual
-//! deltas are shown as points in a graph.
+//! the chosen comparison throughout the whole attempt. Every point of the graph
+//! represents a split. Its x-coordinate is proportional to the split time and
+//! its y-coordinate is proportional to the split delta. The entire diagram is
+//! refered to as the chart and it contains the graph. The x-axis is the
+//! horizontal line that separates positive deltas from negative ones.
+
+// The words "padding" and "content" are from the CSS box model. "Padding" is an
+// area at the top/bottom that stays empty so that the graph doesn't touch the
+// edge. "Content" is the rest (the area inside).
 
 use crate::{
     analysis, comparison,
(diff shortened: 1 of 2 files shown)


KaDiWa4:
The reason the tests are failing:
Previously, the horizontal grid lines in the upper half were 1/120 higher and I think the x-axis was rendered as two lines (whereas it's only rendered once now).

I'll replace the expected hash (with `SXfHSWVpRlc=`).

src/component/graph.rs
```diff
@@ -299,399 +333,281 @@ impl Component {
...
+fn calculate_vertical_scaling(draw_info: &mut DrawInfo) {
+    const MIN_PADDING: f32 = HEIGHT / 24.0;
+    const MAX_CONTENT_HEIGHT: f32 = HEIGHT - MIN_PADDING * 2.0;
+    const SMOOTHNESS: f32 = 0.2;
+
+    let total_delta = draw_info.max_delta - draw_info.min_delta;
+    if total_delta > 0.0 {
+        // A hyperbola works well, this looks something like f(x) = 1/(x + 2)
+        draw_info.padding_y =
+            MAX_CONTENT_HEIGHT * SMOOTHNESS / (total_delta + SMOOTHNESS * 2.0) + MIN_PADDING;
```
wooferzfg, line 513: Is this equivalent to what the old code was doing? Also, could you add a comment explaining what `SMOOTHNESS` is?

wooferzfg (commented):
We might still want the x-axis to be two lines so that it's thicker than the rest of the grid lines and so that it's included in both the positive and negative regions? It would be helpful if you could post screenshots of both the old and new graph for both LSO Desktop and LSO on the web.

src/component/graph.rs
KaDiWa4: Yeah, it's equivalent (aside from float rounding errors)

KaDiWa4:
left half = old version
right half = my new version
<https://cdn.discordapp.com/attachments/667816507614691368/924743455362670612/comparison_images.zip>
it's symmetric now but a bit less visible

PR #494, opened 2021-12-26
Update function names from dependency `heck`

KaDiWa4:
#488 broke `capi/bind_gen` because `heck` renamed its traits/functions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -656,6 +656,22 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+  bind_gen:
+    name: Generate bindings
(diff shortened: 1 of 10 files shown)


CryZe:
Mmh, I guess that means we never build it in CI. Do you feel like adding it to CI somewhere?

CryZe:
There's also some formatting diffs: https://github.com/LiveSplit/livesplit-core/runs/4637782605?check_suite_focus=true

Can you run cargo fmt?

KaDiWa4:
I've never done CI. Also, is there a way to combine those two commits into one?

CryZe:
You can run `cargo reset master` at which point it will undo the commits, but not the changes and then you can do a new commit. That's probably the easiest way. Another is to rebase, but that's a bit more involved.

PR #503, opened 2022-01-08
Support Auto Splitting in the C API

DarkRTA:
note to cryze: do not merge yet


capi/bind_gen/src/c.rs
```diff
@@ -16,7 +16,7 @@ fn get_type(ty: &Type) -> Cow<'_, str> {
         "u32" => "uint32_t",
         "u64" => "uint64_t",
         "usize" => "size_t",
-        "isize" => "ssize_t",
+        "isize" => "ptrdiff_t",
```
KaDiWa4, line 19: You can do the same change in `capi/bind_gen/src/swift/header.rs`.

capi/src/auto_splitting_runtime.rs
```diff
@@ -0,0 +1,76 @@
...
+    let path = if !path.is_empty() {
+        Some(PathBuf::from(path))
+    } else {
+        None
+    };
+
+    match path {
+        Some(x) => this.load_script(x).is_ok(),
+        None => false,
+    }
```
CryZe, line 64: Why build an option just to match on it again? Also you don't need this PathBuf at all. I'm pretty sure you can just pass the &str directly.

capi/bind_gen/src/c.rs
CryZe: Yeah please apply it there too.

DarkRTA:
fixed both of the concerns

*should* be good to merge unless you guys see something else

capi/src/auto_splitting_runtime.rs
DarkRTA: you can't pass in the &str directly otherwise rust complains about mismatched types

capi/src/auto_splitting_runtime.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// Attempts to load an auto splitter. Returns true if successful.
+#[no_mangle]
+pub unsafe extern "C" fn AutoSplittingRuntime_load_script(
+    this: &AutoSplittingRuntime,
+    path: *const c_char,
+) -> bool {
+    let path = str(path);
+    if !path.is_empty() {
+        this.load_script(PathBuf::from(path)).is_ok()
```
CryZe, line 56: You still don't need PathBuf... unless you actually do, then the bug would be in the Runtime. Oh and definitely in your dummy load_script.

CryZe: Oh wait, it's because we send it to a thread. Don't mind me then.

CryZe (approved):

PR #504, opened 2022-01-09
Upgrade `snafu` to 0.7.0

KaDiWa4:


CryZe:
Oof, this is really ugly. I'm not sure if I like this. It seems like there might be a way to turn off this suffix? I think I might prefer that.

KaDiWa4:
There is `#[snafu(context(suffix(false)))]`

CryZe (approved):
I think this looks much better. Though I think it's a bit sad they went this way. Thank you for doing this. 👍

PR #505, opened 2022-01-12
Fix not equal comparison in CI script

CryZe:
Apparently `-ne` is only meant for numbers.


CryZe:
Blocked by: https://github.com/rust-lang/rust/issues/92755

PR #511, opened 2022-01-18
Update sysinfo requirement to 0.23.0

KaDiWa4:


CryZe (changes requested):

CryZe (approved):
Thanks, this looks very good. 👍

PR #512, opened 2022-01-19
Add a function which sets up the "Console Error Panic Hook" so thatg debugging is easier

Hurricane996:


capi/Cargo.toml
```diff
@@ -8,10 +8,11 @@ edition = "2018"
...
 serde_json = "1.0.8"
 time = { version = "0.3.4" }
+console_error_panic_hook = { version = "0.1.7", optional = true }
 
 [features]
 default = ["image-shrinking"]
 image-shrinking = ["livesplit-core/image-shrinking"]
 software-rendering = ["livesplit-core/software-rendering"]
-wasm-web = ["livesplit-core/wasm-web"]
+wasm-web = ["livesplit-core/wasm-web", "console_error_panic_hook"]
```
CryZe, line 17: I think this could be its own features. I'm not sure if you always want this.

PR #521, opened 2022-03-07
fix: fixed an issue with linux hotkey library

jewlexx:


CryZe:
It seems like this is a bug in the evdev library. I opened an issue.

CryZe:
It looks like they yanked the version, so it should compile again now.

jewlexx:
Hey sorry for the late response, I'll test it as soon as I can and let you know

jewlexx:
@CryZe 

It is still not compiling.
Here is my error message if it helps

```
error[E0599]: no method named `supported_events` found for reference `&(PathBuf, Device)` in the current scope
   --> /home/USER/.cargo/registry/src/github.com-1ecc6299db9ec823/livesplit-hotkey-0.6.0/src/linux/mod.rs:281:27
    |
281 |             .filter(|d| d.supported_events().contains(EventType::KEY))
    |                           ^^^^^^^^^^^^^^^^ method not found in `&(PathBuf, Device)`

error[E0277]: a value of type `Vec<Device>` cannot be built from an iterator over elements of type `(PathBuf, Device)`
    --> /home/USER/cargo/registry/src/github.com-1ecc6299db9ec823/livesplit-hotkey-0.6.0/src/linux/mod.rs:282:14
     |
282  |             .collect();
     |              ^^^^^^^ value of type `Vec<Device>` cannot be built from `std::iter::Iterator<Item=(PathBuf, Device)>`
     |
     = help: the trait `FromIterator<(PathBuf, Device)>` is not implemented for `Vec<Device>`
note: required by a bound in `collect
```

CryZe:
You may need to delete the Cargo.lock or something?

jewlexx:
Yep that fixed it thanks

PR #530, opened 2022-05-18
The scope of the unsafe block can be appropriately reduced

peamaeq:


CryZe:
This is superseded by #532

PR #537, opened 2022-06-05
Rainbow best segments

Hurricane996:


Hurricane996:
NOTE: I introduce a function visual_timer_cycle into this.  I think it's conceptually sound (a global timer used to allow animation without state), and the implementation is fine, but I would like someone to majke sure the API makes sense

PR #539, opened 2022-06-06
Timer delta background 

### LINKED ISSUE #380: Implement Background based on Delta for the Timer
I completely forgot that this is an actual feature. I noticed this when I tried to import some layout file that didn't parse correctly.

https://github.com/LiveSplit/LiveSplit.Timer/commit/fbacda12e72435c4793c152fbaec98ee5b100356

Hurricane996:
Fixes #380


CryZe:
I fixed the cross building on master, so you may want to rebase your changes. Though there seem to be some test failures regardless.

CryZe:
Seems like the rendering tests still fail. I'm guessing (a few of) the layouts are using the setting and now look different because of it.

PR #589, opened 2022-10-31
Changes Process.with_name to sort pids by id

Eein:
Prior to selecting max key by start time, this change sorts the keys to ascending to take the higher PID in case of all records having an equal `start_time()`

see: https://doc.rust-lang.org/nightly/src/core/iter/traits/iterator.rs.html#3011-3012

This is to fix a case where attempting to attach to steam proton processes on linux may return two PIDs with the same start_time, potentially selecting the wrong pid until a reload happens with the correct pid.

---

Apologies for the lack of tests - I couldn't figure out how to mock things in rust as I dont have access to sysinfo::Process fields

Note: includes a `cargo fmt`

Tested with a custom build of `obs-livesplit-one` and it seems to work perfectly - Have not tested windows; but the obs plugin will need an update.

### DIFF
--- a/crates/livesplit-auto-splitting/src/process.rs
+++ b/crates/livesplit-auto-splitting/src/process.rs
@@ -44,9 +44,17 @@ impl Process {
         process_list.refresh();
         let processes = process_list.processes_by_name(name);
 
-        // Use the process that was started the most recently, it's more
+        // Sorts the processes (asc) by numeric pid, to allow max_by_key to
+        // select the higher pid in case all records are equally maximum; otherwise
+        // use the process that was started the most recently, it's more
         // predictable for the user.
-        let pid = processes
+
+        let mut procs: Vec<&sysinfo::Process> = processes.map(|p| p).collect();
+
+        procs.sort_unstable_by_key(|p| p.pid().as_u32());
+
+        let pid = procs
+            .iter()
             .max_by_key(|p| p.start_time())
             .context(ProcessDoesntExist)?
             .pid()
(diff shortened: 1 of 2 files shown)


crates/livesplit-auto-splitting/src/process.rs
```diff
@@ -44,9 +44,17 @@ impl Process {
...
-        // Use the process that was started the most recently, it's more
+        // Sorts the processes (asc) by numeric pid, to allow max_by_key to
+        // select the higher pid in case all records are equally maximum; otherwise
+        // use the process that was started the most recently, it's more
         // predictable for the user.
-        let pid = processes
+
+        let mut procs: Vec<&sysinfo::Process> = processes.map(|p| p).collect();
+
+        procs.sort_unstable_by_key(|p| p.pid().as_u32());
```
CryZe, line 54: Considering we only need the maximum here, we don't need to fully sort everything. Instead you can use a tuple in `max_by_key` made up of `(start_time, pid)`. That way it sorts primarily by `start_time` but falls back to the pid if they are equal.

CryZe (changes requested):

crates/livesplit-auto-splitting/src/process.rs
Eein: Thank you; A big light bulb went off as i was trying to do something similar.
Since Tuple implements `Ord` - it knows how to find the max value, correct?

Eein:
I also tested this - seems to work just fine.

CryZe (approved):

PR #591, opened 2022-11-04
Implement `SpeedRunIGT` Parsing

CryZe:
`SpeedRunIGT` is a Minecraft mod that brings a full timer with deep auto splitting capabilities to Minecraft. They recently added support for https://therun.gg and it seems like it's quite commonly used, probably the most commonly used timer after LiveSplit at this point. So it makes sense for us to be able to parse these splits.

I also snuck in a few cleanups along the way (some now possible because of Rust 1.65).

cc @RedLime

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -126,7 +126,6 @@ std = [
     "serde/std",
     "simdutf8/std",
     "snafu/std",
-    "time/formatting",
     "time/local-offset",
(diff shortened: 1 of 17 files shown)


CryZe:
Looks like this is blocked on https://github.com/Lokathor/bytemuck/pull/140

Issue #372, opened 2020-11-19
Add links to types in documentation


CryZe:
Rust 1.48 just came out with a proper built-in way to mention types in documentation that you can click on and get linked to that specific type. Before you needed to write the relative URL of the type which was both cumbersome and based on an implementation detail, so we just never linked to any types. Here's the blog post:

[Rust 1.48](https://blog.rust-lang.org/2020/11/19/Rust-1.48.html)

Eein:
I've started a little work on this. I've attempted to avoid changing as much of the wording as possible for now; however I feel like some things are looking to busy now: See:

![image](https://user-images.githubusercontent.com/5025835/200201296-f59717aa-2505-4419-8720-8b0df18b9310.png)

CryZe:
Yeah that sounds fine.

PR #596, opened 2022-11-08
Adds types links to docs

Eein:
#372 

This PR adds type links to the rust docs; It tries to keep code formatting similarly to what was there.


CryZe:
Thanks this is great! 🎉

PR #600, opened 2022-11-10
Optimize accessing PNG dimensions

CryZe:
I've noticed that a lot of the parsing is spent figuring out the dimensions of our icons, which very often are PNG images. Apparently to access the full header of a PNG file a lot of decoding needs to happen, which is apparently quite expensive. However we don't really need to access the full header and instead really only the width and height. And here the PNG specification comes in handy:

https://www.w3.org/TR/2003/REC-PNG-20031110/

It says the following:

> A valid PNG datastream shall begin with a PNG signature, immediately followed by an IHDR chunk

Each chunk is encoded as a length and type and then its chunk encoding. An IHDR chunk immediately starts with the width and height. This means we can model the beginning of a PNG file as follows:

```rust
// 5.2 PNG signature
png_signature: [u8; 8],
// 5.3 Chunk layout
chunk_len: [u8; 4],
// 11.2.2 IHDR Image header
chunk_type: [u8; 4],
width: U32,
height: U32,
```

And with a little bit of bytemuck magic we essentially can access the width and height with a single read. This improves parsing speed of entire splits files by up to 30%. Yes, the old code was that slow.

### DIFF
--- a/src/run/parser/livesplit.rs
+++ b/src/run/parser/livesplit.rs
@@ -433,7 +433,7 @@ fn parse_attempt_history(version: Version, reader: &mut Reader<'_>, run: &mut Ru
 pub fn parse(source: &str, path: Option<PathBuf>) -> Result<Run> {
     let reader = &mut Reader::new(source);
 
-    let mut image_buf = Vec::with_capacity(4096);
+    let mut image_buf = Vec::new();
 
     let mut run = Run::new();
 
--- a/src/settings/image/shrinking.rs
+++ b/src/settings/image/shrinking.rs
@@ -1,26 +1,62 @@
 use alloc::borrow::Cow;
+use bytemuck::{Pod, Zeroable};
 use image::{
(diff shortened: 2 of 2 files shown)


src/run/parser/livesplit.rs
```diff
@@ -433,7 +433,7 @@ fn parse_attempt_history(version: Version, reader: &mut Reader<'_>, run: &mut Ru
 pub fn parse(source: &str, path: Option<PathBuf>) -> Result<Run> {
     let reader = &mut Reader::new(source);
 
-    let mut image_buf = Vec::with_capacity(4096);
+    let mut image_buf = Vec::new();
```
CryZe, line 436: I also snuck this in. Nowadays we know exactly how much memory we need for each image in here when we encounter it, so there's no need to preallocate anything if we are going to reallocate the memory anyway the moment we encounter an image.

PR #615, opened 2022-12-16
fix title abbreviations

DarkRTA:


crates/livesplit-title-abbreviations/src/lib.rs
```diff
@@ -103,7 +103,7 @@ fn is_all_caps_or_digits(text: &str) -> bool {
 
 pub fn abbreviate(name: &str) -> Vec<Box<str>> {
     let name = name.trim();
-    let mut list = vec![name.into()];
+    let mut list = vec![];
     if name.is_empty() {
         return list;
     }
```
CryZe, line 109: This now returns an empty list instead of one containing the name.

CryZe (changes requested):

crates/livesplit-title-abbreviations/src/lib.rs
DarkRTA: i just tested this function and came across the opposite problem,
the list contains all of the abbreviations but there's a duplicate that now needs to be accounted for

CryZe: Maybe you could add some tests then, if it's so easy to overlook these problems xD

DarkRTA: you know i am probably going to do that

crates/livesplit-title-abbreviations/src/lib.rs
```diff
@@ -165,6 +165,12 @@ pub fn abbreviate(name: &str) -> Vec<Box<str>> {
     list.sort_unstable();
     list.dedup();
 
+    if let Some(idx) = list.iter().position(|x| name == x.as_ref()) {
+        list.remove(idx);
```
DarkRTA, line 169: rationale: we need to remove the full name from the list if it already exists, and at the same time i'd like to preserve the sorting

if speed is desired over sorting, this can be changed to a `Vec::swap()`

crates/livesplit-title-abbreviations/src/lib.rs
```diff
@@ -214,3 +220,71 @@ pub fn abbreviate_category(category: &str) -> Vec<Box<str>> {
...
+
+        assert_eq!(abbreviations, expected);
+    }
+    #[test]
+    fn sm64() {
+        let abbreviations = abbreviate("Super Mario 64");
+
+        let expected = vec![
+            Box::from("SM64"),
+            Box::from("Super Mario 64")
```
DarkRTA, line 285: apparently `rustfmt` wants these to be on a single line, which arguably looks worse imo

CryZe: There's #[rustfmt::skip] you can put, if you really don't want to take its suggestion.

crates/livesplit-title-abbreviations/src/lib.rs
CryZe: Yeah `swap` almost definitely makes more sense as there is no guaranteed order in here other than the last element.

CryZe (approved):

PR #627, opened 2023-01-24
Implement Linked Layouts

CryZe:
This implements linked layouts that allow you to associate splits with layouts. This means that a timer application can automatically open the layout that's meant to be used for a particular splits file, improving the user experience a lot. This already landed in the original LiveSplit, so while the way it is stored in the XML is not ideal, in particular with the way the default layout is annotated, we will have to support it for backwards compatibility anyway. Further changes can be made in the future.

cc @Relacibo

Continuation of #436

### DIFF
--- a/capi/src/lib.rs
+++ b/capi/src/lib.rs
@@ -35,6 +35,7 @@ pub mod layout;
 pub mod layout_editor;
 pub mod layout_editor_state;
 pub mod layout_state;
+pub mod linked_layout;
 pub mod parse_run_result;
 pub mod pb_chance_component;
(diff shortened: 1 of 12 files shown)


wooferzfg:
In this PR, https://github.com/LiveSplit/livesplit-core/pull/627/commits/cf696559d431f01ad30ed127493c83903a79b4c9 matches the commit we're using in LiveSplit.

CryZe:
Oh, so `<None />` as it is in the latest version, isn't actually a thing in what we used in LiveSplit?

PR #634, opened 2023-02-14
livesplit-hotkey: Make `Hook` Send + Sync

DarkRTA:


DarkRTA:
Wow, apparently the static assertion actually caught some instaces where `Hook` is not `Send + Sync` on other platforms.

Issue #623, opened 2023-01-15
Ask for Permission for the Hotkeys on Mac


savage13:
I made an attempt to get Hotkeys working on Mac.  The call to `CGEventTapCreate()` returns null as it does not have permission.

https://github.com/LiveSplit/livesplit-core/blob/fc9a1c92fb5099793ebf250d241361f5f83a9ac1/crates/livesplit-hotkey/src/macos/mod.rs#L112

You can ask the system to show the Security and Privacy preferences (in System Preferences) by following
- https://stackoverflow.com/a/24100470 or 
- https://developer.apple.com/forums/thread/109283 (working C program)

Once the system is granted permission, the HotKeys appear to work properly.

To call these mac functions from rust requires a set of definitions for CFDictionary and friends added to

https://github.com/LiveSplit/livesplit-core/blob/fc9a1c92fb5099793ebf250d241361f5f83a9ac1/crates/livesplit-hotkey/src/macos/cg.rs

PR #684, opened 2023-06-14
macOS permission module

lbrndnr:
Hi,

I have added a `permission` module that makes it possible to check and ask for accessibility permissions, as discussed in #623. I figured it would be nice to put this functionality in a new module, but let me know if you have another opinion. I'm relatively new to Rust so maybe double check my code :)


CryZe:
Thanks a lot, this is great :)

I made some adjustments so it automatically happens, which should be fine as it simply skips the popup if the permission is already granted.

PR #713, opened 2023-08-20
auto_splitting feature: Add requests for auto splitter settings management

Refragg:
This patch adds 4 requests to the auto-splitting feature so that the library users are able to manage the auto splitter settings:
- GetSettings: Gets all of the user settings that are meant to be shown to the runner
- GetSettingsValue(key: String): Gets the SettingValue of a specific setting
- SetSettingsValue(key: String, value: SettingValue): Sets the SettingValue of a specific setting
- ReloadScript: Helper request that will reload the current script while keeping the same SettingsStore

These new requests are part of my ongoing work towards having the settings editable in the obs-livesplit-one UI, I have tested them to work as expected.

I don't know why the Windows i686 gnu build fails but it doesn't seem like it's related to this pull request?

### DIFF
--- a/crates/livesplit-auto-splitting/src/runtime.rs
+++ b/crates/livesplit-auto-splitting/src/runtime.rs
@@ -280,6 +280,11 @@ impl<T: Timer> Runtime<T> {
         &self.store.data().settings_store
     }
 
+    /// Accesses the currently stored settings as mutable.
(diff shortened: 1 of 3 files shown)


CryZe:
Ah yes, the CI is very green. Let's merge it.

PR #726, opened 2023-10-02
WIP: Add get_auto_splitter_settings for AutoSplitterSettings XML contents

AlexKnauth:
This adds `get_auto_splitter_settings` to the Auto Splitting Runtime, to get the `AutoSplitterSettings` XML contents.

There may be better ways to accomplish this, and I may run into problems when trying to use it in an ASR-based autosplitter, so this is a draft until I know an autosplitter can make good use of it.

### DIFF
--- a/crates/livesplit-auto-splitting/README.md
+++ b/crates/livesplit-auto-splitting/README.md
@@ -226,6 +226,15 @@ extern "C" {
         tooltip_ptr: *const u8,
         tooltip_len: usize,
     );
+    /// Stores the AutoSplitterSettings XML contents in the
+    /// buffer given.
+    /// Returns `false` if the buffer is too small.
+    /// After this call, whether it was successful or not,
+    /// the `buf_len_ptr` will be set to the required buffer size.
+    pub fn get_auto_splitter_settings(
+        buf_ptr: *mut u8,
+        buf_len_ptr: *mut usize,
+    ) -> bool;
 }
 ```
 
--- a/crates/livesplit-auto-splitting/src/lib.rs
+++ b/crates/livesplit-auto-splitting/src/lib.rs
@@ -226,6 +226,15 @@
 //!         tooltip_ptr: *const u8,
 //!         tooltip_len: usize,
 //!     );
+//!     /// Stores the AutoSplitterSettings XML contents in the
+//!     /// buffer given.
+//!     /// Returns `false` if the buffer is too small.
+//!     /// After this call, whether it was successful or not,
+//!     /// the `buf_len_ptr` will be set to the required buffer size.
+//!     pub fn get_auto_splitter_settings(
+//!         buf_ptr: *mut u8,
+//!         buf_len_ptr: *mut usize,
+//!     ) -> bool;
 //! }
 //! ```
 //!
--- a/crates/livesplit-auto-splitting/src/runtime.rs
+++ b/crates/livesplit-auto-splitting/src/runtime.rs
@@ -980,6 +980,25 @@ fn bind_interface<T: Timer>(linker: &mut Linker<Context<T>>) -> Result<(), Creat
         .map_err(|source| CreationError::LinkFunction {
             source,
             name: "user_settings_set_tooltip",
+        })?
+        .func_wrap("env", "get_auto_splitter_settings", {
+            |mut caller: Caller<'_, Context<T>>, ptr: u32, len_ptr: u32| {
+                let (memory, context) = memory_and_context(&mut caller);
+                let auto_splitter_settings = &context.settings_store.get_auto_splitter_settings();
+                let len_bytes = get_arr_mut(memory, len_ptr)?;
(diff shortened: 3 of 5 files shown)


CryZe:
I unfortunately should've said this earlier, or worded it more strongly. I don't think we'll merge this because XML probably isn't going to be the way these settings are stored in the long term. XML was a great choice 10 years ago, but nowadays XML is just way too verbose. So I would eventually probably like to move our splits file to another format, maybe one that is more compact and parses even faster. This PR would tie us down to a specific format, which would go against that entire idea. I also don't see why you would need this. There's be a whole querying / modifying API for the settings. You shouldn't need to parse the entire settings "file". You can however store arbitrary strings in there, so you could encode a JSON or XML or whatever into a value. That's always possible.

AlexKnauth:
There are existing splits files, in the existing XML format, and I want my autosplitter to support those existing splits files.

When a better settings store querying API is added in the future, how will it support splits files in the old format?

AlexKnauth:
If the `XML` looks like this:
```xml
    <Ordered>True</Ordered>
    <AutosplitEndRuns>True</AutosplitEndRuns>
    <AutosplitStartRuns/>
    <Splits>
      <Split>KingsPass</Split>
      <Split>Grub1</Split>
      <Split>EnterBroodingMawlek</Split>
      <Split>MaskFragment1</Split>
      <Split>Grub2</Split>
      <Split>MaskFragment2</Split>
      <Split>GruzMother</Split>
      <Split>SlyRescued</Split>
      <Split>Grub3</Split>
      <Split>Grub4</Split>
      <Split>Grub5</Split>
      <Split>MaskFragment3</Split>
      <Split>Mask1</Split>
    </Splits>
```
And a translated JSON version, for the sake of example, could look like this:
```json
{ 
  "Ordered": true,
  "AutosplitEndRuns": true,
  "AutosplitStartRuns": "",
  "Splits": [
    { "Split": "KingsPass" },
    { "Split": "Grub1" },
    { "Split": "EnterBroodingMawlek" },
    { "Split": "MaskFragment1" },
    { "Split": "Grub2" },
    { "Split": "MaskFragment2" },
    { "Split": "GruzMother" },
    { "Split": "SlyRescued" },
    { "Split": "Grub3" },
    { "Split": "Grub4" },
    { "Split": "Grub5" },
    { "Split": "MaskFragment3" },
    { "Split": "Mask1" },
  ],
}
```
A good querying API would be able to extract the same information from both formats, I guess. But there would have to be some translation defined between them.

CryZe:
Yeah that's roughly how it would work. I'll focus on this querying / modification API and settings reloading without restarting the auto splitter next. I think most of these problems will go away once this API exists.

AlexKnauth:
With the similarities in the structure of the XML and JSON versions above, I can imagine these ways of querying:

Query as a boolean: For XML, querying `True` as a boolean is true, `False` is false. JSON booleans are just booleans.

Query as a string: For XML, querying `X`, as a string with no tags in it, gives you `"X"`. JSON strings are just strings.

Query as a dictionary mapping keys to values: For XML, the keys are tags, like querying `B` in `<A>X</A><B>Y</B><C>Z</C>` could give you a query-able object value for `Y`. For JSON, the keys are JSON object fields, like querying `B` in `{ "A": X, "B": Y, "C", Z }` could give you a query-able object value for `Y`.

Query as a list: For XML, querying `<A>X</A><B>Y</B><C>Z</C>` as a list gives you 3 elements, query-able objects corresponding to `<A>X</A>`, `<B>Y</B>`, and `<C>Z</C>` respectively. To get the `X`, `Y`, and `Z` values from this you'd have to query each element as a dictionary so `.list_get(0).dict_get("A")` produces a query-able object for `X`, and so on.

So with either of the XML or JSON versions above as a query-able object named `obj`, a query to get the string `"EnterBroodingMawlek"` from that structure could look like:
```
obj.dict_get("Splits").list_get(2).dict_get("Split").string_get()
```

AlexKnauth:
Here's a trait with the query functions implemented on XML:
https://github.com/AlexKnauth/wasm-hollowknight-autosplit-prototype/blob/auto_splitter_settings/src/auto_splitter_settings.rs
```rust
pub trait Settings: Clone {
    fn as_str(&self) -> Option<&str>;
    fn as_bool(&self) -> Option<bool>;
    fn as_list(&self) -> Option<Vec<Self>>;
    fn as_dict(&self) -> Option<BTreeMap<String, Self>>;

    fn list_get(&self, index: usize) -> Option<Self> {
        self.as_list()?.get(index).cloned()
    }
    fn dict_get(&self, key: &str) -> Option<Self> {
        self.as_dict()?.get(key).cloned()
    }
}
```

If the XML can be abstracted away into a query-able object as flexible as this, that would be enough. But I don't yet know enough about the `extern C` and FFI stuff to implement this instead of a function that just gets a string that's already there.

PR #728, opened 2023-10-07
Add `optimize` setting to Auto Splitting Runtime

CryZe:
It turns out that when debugging, you really don't want to optimize the WebAssembly module as it loses a lot of information about the variables. This adds a setting to make it configurable.


CryZe:
Yay more broken targets.

PR #735, opened 2023-10-31
Add process_get_module_path

AlexKnauth:
This adds `process_get_module_path` to the Auto Splitting Runtime, which gets the path on the file system to the module.

This is useful in situations where the Section Header Table is there in the file, but is not loaded into memory, even when the rest of the module is.


CryZe:
This looks good. I'd like to improve the docs a bit. So I'll rebase this for you.

AlexKnauth:
I also saw the changes to `process_get_path` in #736, and I would like to include analogs of those changes in here.

CryZe:
Yeah I'll port those as well.

PR #754, opened 2024-01-02
Allowing the runtime to correctly hook into long-named Windows processes on Linux

Jujstme:


CryZe:
There's a few issues:
1. I think I'd prefer a #[cfg] or cfg! based approach, as it more cleanly compiles away.
2. `str` always has to be valid UTF-8 in Rust, meaning that you can only cut, if you know that you don't break the UTF-8. So if the process name contains Emojis or so, and they get destroyed... The program would panic. I'd be curious to know what even happens to the process name (that we get from Linux) in that case then, i.e. what happens in the crate that we use for getting the name. Does it also just panic? We definitely need to put a bunch of emojis in a process name and test it.
3. Is this the approach we even want? Maybe if the process name is too long, we also ensure by matching the path?

CryZe:
Actually nvm, you cut the length provided from the auto splitter since the second commit. This of course makes it so the UTF-8 validation errors out there and the auto splitter gets turned off instead. That's definitely safer, but I wonder if we can just compare the based instead, i.e. half an emoji is fine to compare too.

Jujstme:
The linux kernel allows a buffer of 16 bytes for the process name (including the null termination), which means essentially 15 bytes. If a longer buffer is provided, it just gets truncated. This means possible Emojis or so in the process name should not be a problem.

For the #[cfg], I can restore it easily.

CryZe:
It is a problem because Rust doesn't allow invalid UTF-8, no matter what Linux does. So we need to figure out what happens with the process name we get from the upstream crate in order to properly cut and compare it.

Jujstme:
The `get_str()` function in the runtime internally calls `str::from_utf8()`, so if the slice is not a valid UTF-8 the `attach()` function should just silently fail.

Issue #710, opened 2023-07-30
[livesplit-hotkey] Ability to prevent hotkeys from propagating to other programs


Xevion:
Hello - I am looking for a crate with the ability to (on Windows) add a global hook for listening to specific hotkeys (like `WIN+ALT+LEFT|RIGHT`).
I want to be able to prevent these hotkeys from propagating to whatever program is currently selected.
Which is problem, as `livesplit-hotkey` does not do that - the hotkeys are still received by whatever program is currently selected.

Researching the matter, I'm really unsure of how the Windows APIs work, if what I want is even possible, and whether LiveSplit might ever carry the feature in the future. Given the nature of your app, I get the feeling the answer is 'no', but I might hope that you'll have some small insight available for how I could proceed.

I'm fine with you closing this issue immediately - I'm only hoping for some assistance since I have a really close MVP based on `livesplit-hotkey`; I don't want to be re-engineering an entire hotkey library for the sole purpose of preventing hotkey propagation.

- I am only looking for a Windows solution - other platforms need not bother.
- My intended program will ONLY be listening to hotkeys - few/no windows.

PR #755, opened 2024-01-14
livesplit-hotkey: Add Hook::new_consuming on macOS

tmandry:
This was also requested in #710 for Windows, so there is some demand for it, though I recognize it diverges from the crate's original use case. If it's not something you want to support I can make a bespoke implementation elsewhere, but this happened to be the cleanest implementation/API I could find of a global shortcut library for macOS.

The API puts the choice of consuming or not on the Hook because it allows us to mark non-consuming hooks listen-only, which removes them as a potential bottleneck from the system input pipeline. I think this is probably desirable for your use case but I'm not certain.

After making this change I verified locally that both `new` and `new_consuming` work as expected, on macOS 13.5.2.

### DIFF
--- a/crates/livesplit-hotkey/src/macos/mod.rs
+++ b/crates/livesplit-hotkey/src/macos/mod.rs
@@ -21,6 +21,7 @@ use self::{
     },
 };
 use crate::{ConsumePreference, Hotkey, KeyCode, Modifiers, Result};
+use core::ptr::null_mut;
 use std::{
     collections::{hash_map::Entry, HashMap},
     ffi::c_void,
@@ -97,9 +98,10 @@ impl Drop for Hook {
 
 impl Hook {
     pub fn new(consume: ConsumePreference) -> Result<Self> {
-        if matches!(consume, ConsumePreference::MustConsume) {
-            return Err(crate::Error::UnmatchedPreference);
-        }
+        let is_consuming = matches!(
+            consume,
+            ConsumePreference::PreferConsume | ConsumePreference::MustConsume,
+        );
 
         let state = Arc::new(State {
             hotkeys: Mutex::new(HashMap::new()),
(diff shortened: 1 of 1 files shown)


crates/livesplit-hotkey/src/macos/mod.rs
```diff
@@ -488,6 +502,7 @@ unsafe extern "C" fn callback(
         .get_mut(&key_code.with_modifiers(modifiers))
     {
         callback();
+        CGEventSetType(event, EventType::NULL);
```
CryZe, line 505: This seems to be the main change, considering we actually (accidentally?) had `DEFAULT_TOP` in the non-consuming implementation. Apple's documentation seems very unclear that this would cause the key to be consumed. Does this not need any `if` that guards it to only one implementation? No matter what, I'd like some comment explaining what's going on here.

CryZe (changes requested):
The original LiveSplit actually had a setting for this, so it would make sense for this crate to support this feature as well. We probably can't make it work on all platforms, but we could always fall back to the other implementation if necessary. I'll push some more changes that at least ensures that every platform has the same API later today.

CryZe:
Alright, I rebased your changes onto the new API introduced in #757

crates/livesplit-hotkey/src/macos/mod.rs
tmandry: I've seen some implementations do this in the wild, but after reviewing [this page](https://developer.apple.com/documentation/coregraphics/cgeventtapcallback) I realized it isn't necessary and we can just return a null pointer. That clearly says that the return value will be ignored in the passive listener case. Yes `DEFAULT_TAP` could have been `LISTEN_ONLY` before, but now the difference will become load bearing. I added a comment.

CryZe (approved):
Thank you this looks good now.

PR #758, opened 2024-01-17
pub use livesplit_auto_splitting::wasi_path

AlexKnauth:
Re-exports `wasi_path` from `livesplit_core::auto_splitting` so that https://github.com/LiveSplit/obs-livesplit-one/pull/47 can use it without adding an additional dependency.

### DIFF
--- a/src/auto_splitting/mod.rs
+++ b/src/auto_splitting/mod.rs
@@ -544,8 +544,7 @@ use crate::{
     platform::Arc,
     timing::{SharedTimer, TimerPhase},
 };
-pub use livesplit_auto_splitting::settings;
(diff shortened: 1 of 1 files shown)


src/auto_splitting/mod.rs
```diff
@@ -545,6 +545,7 @@ use crate::{
     timing::{SharedTimer, TimerPhase},
 };
 pub use livesplit_auto_splitting::settings;
+pub use livesplit_auto_splitting::wasi_path;
```
CryZe, line 548:
```suggestion
pub use livesplit_auto_splitting::{settings, wasi_path};
```

PR #790, opened 2024-04-11
Update resource paths

just-ero:
Parent PR: https://github.com/LiveSplit/LiveSplit/pull/2446

### Description

Some resources were moved on the main `LiveSplit/LiveSplit` repo which affects some README files here.


CryZe:
These should've referenced the SVGs, like in the main LiveSplit README.

wooferzfg:
https://github.com/LiveSplit/livesplit-core/pull/791

PR #797, opened 2024-04-30
Run parser: deserialize the AutoSplitterSettings

AlexKnauth:
Based on the work of @Refragg, I updated it to use the new settings map and settings value types.

- [x] Make sure the auto-splitter gets the settings from a freshly opened splits file.
- [x] Make sure a saved splits file gets the settings from the autosplitter, when the autosplitter has edited them.
- [x] Make sure a saved splits file gets to keep its Legacy XML, when the autosplitter has no settings of its own to replace it with.

### DIFF
--- a/crates/livesplit-auto-splitting/src/settings/gui.rs
+++ b/crates/livesplit-auto-splitting/src/settings/gui.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 /// A setting widget that is meant to be shown to and modified by the user.
 #[non_exhaustive]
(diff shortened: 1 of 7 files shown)


AlexKnauth:
With how I'm looking at this again now after seeing how an ASR autospliter's settings work slightly differently once it's actually in the `LiveSplit.AutoSplitters.xml`, I think I want to take out the `ScriptPath` element from this PR... it wasn't really doing anything anyway

CryZe:
#943 is now merged.

PR #806, opened 2024-05-22
layout setting: Running Ignore Mouse

AlexKnauth:
Add a layout setting for Ignore Mouse While Running and Not In Focus.

### DIFF
--- a/src/layout/general_settings.rs
+++ b/src/layout/general_settings.rs
@@ -50,6 +50,8 @@ pub struct GeneralSettings {
     pub separators_color: Color,
     /// The text color to use for text that doesn't specify its own color.
     pub text_color: Color,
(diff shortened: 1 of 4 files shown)


CryZe:
In LSO layouts are a bit different in that their purpose is only for rendering, they don't handle any interactions, auto splitting, manual game time or other platform specific stuff. They truly only visualize information. Similar settings like "Always on Top" for example are in the general settings for the specific frontend, if it can even support that. We likely want to have a way to override the general settings for specific games / splits anyway, so you can have different hotkeys for different games and stuff. So I don't think the layout is the place anymore in LSO for this setting.

CryZe:
I think when parsing from a `.lsl` we can return not only a `Layout` but also some sort of `AdditionalLslSettings` that you could optionally apply.

PR #809, opened 2024-06-01
Add a timeout message

AlexKnauth:
To distinguish between errors where an auto-splitter update took too long, from other errors, such as panics, running out of stack space, etc.

### DIFF
--- a/src/auto_splitting/mod.rs
+++ b/src/auto_splitting/mod.rs
@@ -952,6 +952,8 @@ fn watchdog<T: event::CommandSink + TimerQuery + Send>(shared_state: Arc<SharedS
                     if let Some(auto_splitter) = &*shared_state.auto_splitter.load() {
                         auto_splitter.interrupt_handle().interrupt();
                     }
+                } else {
+                    has_timed_out = false;
                 }
 
                 new_state
--- a/src/auto_splitting/mod.rs
+++ b/src/auto_splitting/mod.rs
@@ -920,6 +920,7 @@ fn run<T: event::CommandSink + TimerQuery + Send>(
 
 fn watchdog<T: event::CommandSink + TimerQuery + Send>(shared_state: Arc<SharedState<T>>) {
     const TIMEOUT: Duration = Duration::from_secs(5);
+    let mut has_timed_out = false;
 
     let Ok(mut state) = shared_state.watchdog_state.lock() else {
         return;
@@ -944,6 +945,10 @@ fn watchdog<T: event::CommandSink + TimerQuery + Send>(shared_state: Arc<SharedS
                 };
 
                 if result.timed_out() {
+                    if !has_timed_out {
+                        log::error!(target: "Auto Splitter", "timeout, no update in {} seconds", TIMEOUT.as_secs_f32());
+                        has_timed_out = true;
+                    }
                     if let Some(auto_splitter) = &*shared_state.auto_splitter.load() {
                         auto_splitter.interrupt_handle().interrupt();
                     }


AlexKnauth:
When testing with an auto-splitter that loops forever printing `looping` without calling `next_tick().await`, the logs look like this:
```
[2024-05-31 20:08:11][Auto Splitter][INFO] looping
[2024-05-31 20:08:11][Auto Splitter][INFO] looping
[2024-05-31 20:08:11][Auto Splitter][INFO] looping
[2024-05-31 20:08:11][Auto Splitter][ERROR] timeout, no update in 5 seconds
[2024-05-31 20:08:11][Auto Splitter][INFO] looping
[2024-05-31 20:08:11][Auto Splitter][ERROR] Unloaded, because the script trapped: error while executing at wasm backtrace:
    0:  0xb6f - <unknown>!<wasm function 17>
    1: 0x79c9 - <unknown>!<wasm function 106>

Caused by:
    wasm trap: interrupt
```

AlexKnauth:
Tested again after a rebase on the new watchdog implementation, logs look like this:
```
[2024-11-16 09:33:24][Auto Splitter][INFO] looping
[2024-11-16 09:33:24][Auto Splitter][INFO] looping
[2024-11-16 09:33:24][Auto Splitter][INFO] looping
[2024-11-16 09:33:24][Auto Splitter][ERROR] timeout, no update in 5 seconds
[2024-11-16 09:33:24][Auto Splitter][INFO] looping
[2024-11-16 09:33:24][Auto Splitter][ERROR] Unloaded, because the script trapped: error while executing at wasm backtrace:
    0: 0x647c6 - hollowknight_autosplit_wasm.wasm!free
    1: 0x49f79 - hollowknight_autosplit_wasm.wasm!hollowknight_autosplit_wasm::main::{{closure}}::haf64fc4596a16a8a
    2: 0x4e2e6 - hollowknight_autosplit_wasm.wasm!update
    3: 0x66def - hollowknight_autosplit_wasm.wasm!update.command_export

Caused by:
    wasm trap: interrupt
```
For an auto-splitter that loops forever printing `looping` without calling `next_tick().await`. Printing an error log message saying `timeout, no update in 5 seconds` is more helpful in this situation, rather than leaving it ambiguous about the possible causes of the error.

src/auto_splitting/mod.rs
```diff
@@ -944,6 +945,10 @@ fn watchdog<T: event::CommandSink + TimerQuery + Send>(shared_state: Arc<SharedS
                 };
 
                 if result.timed_out() {
+                    if !has_timed_out {
+                        log::error!(target: "Auto Splitter", "timeout, no update in {} seconds", TIMEOUT.as_secs_f32());
+                        has_timed_out = true;
```
CryZe, line 950: This never gets reset to false again. The rest looks good.

CryZe: Btw, long term I think we want the watchdog to be more configurable (the timeout) and with a custom callback where the UI could then for example ask the user if they want to shut down the auto splitter.

AlexKnauth: Thanks for the catch, fixed in `timeout: reset to false again`.

src/auto_splitting/mod.rs
```diff
@@ -944,9 +945,15 @@ fn watchdog<T: event::CommandSink + TimerQuery + Send>(shared_state: Arc<SharedS
...
+                    }
                     if let Some(auto_splitter) = &*shared_state.auto_splitter.load() {
                         auto_splitter.interrupt_handle().interrupt();
                     }
+                } else {
+                    has_timed_out = false;
```
CryZe, line 956: I'm not sure this is the right place. I think this should be set in the Unloaded case, so it doesn't get set all the time.

CryZe: Nvm, the reload case in the other thread never puts the watchdog into the unloaded state, so I guess this is fine.

AlexKnauth:
So is this good to merge?

PR #813, opened 2024-06-09
Dynamically adjust the column widths

CryZe:
Instead of assuming each column has the width of the text `88:88:88`, we now determine the largest string per column (based on unicode scalar values) and adjust the column width accordingly. There is still a fixed minimum width based on the string `88:88`, which ensures that the columns don't switch sizes too often, especially if you start with empty splits. We could've also chosen `8:88:88`, which would ensure there's no column size change as you approach the hour mark. However, this would increase the size of delta columns unnecessarily. So it's a little bit of a trade-off.

Before:

image

After:

https://github.com/LiveSplit/livesplit-core/assets/1451630/01d2b917-9046-4107-87ed-a45ab24ae911

### DIFF
--- a/src/rendering/component/splits.rs
+++ b/src/rendering/component/splits.rs
@@ -1,5 +1,3 @@
-use core::iter;
-
 use crate::{
     component::splits::State,
     layout::{LayoutDirection, LayoutState},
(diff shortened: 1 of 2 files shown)


CryZe:
The tests changed, because the deltas are now often closer to the split time column.

CryZe:
@wooferzfg I don't really want a code review, but more of a review of the concept, that we look for the largest string per column, but use `88:88` as the minimum size.

Issue #260, opened 2019-10-23
Design a Timer Synchronization Algorithm


CryZe:
In order to visualize LiveSplit One as a Twitch extension we need to synchronize the runner's timer to the browser. Additionally we want to synchronize information between different LiveSplit One's when racing with each other. So we need to design a synchronization algorithm. We want to design it in such a way, that splits i/o is using it too for their racing feature. And they want it to be timer agnostic, so it needs to support "dumb clients" that don't necessarily track everything properly or can even parse any split formats. However proper timer's such as LiveSplit should synchronize pretty much everything without any loss of information. So the protocol needs to be flexible in this regard where each client can specify their capabilities.

Here's some rough ideas of how the protocol could look like (ignoring encoding for now):
```rust
struct ClientCapabilities {
    can_parse_splits_of: Vec<String>,
    wants_splits_for_every_attempt: bool,
    wants_rough_info_for_every_attempt: bool,
}

enum Event {
    StartAttempt(Option<Splits>),
    DiscardAttempt,
    ResetAttempt,
    FinishAttempt,
    UpdateSplit(usize, Time),
}

enum SplitsInfo {
    Splits(Vec<u8>),
    RoughInfo(SplitsInfo),
}

struct SplitsInfo {
    game: String,
    category: String,
    attempts: usize, // This corresponds with the attempt count BEFORE the attempt started.
    segments: Vec<SegmentInfo>,
}

struct SegmentInfo {
    name: String,
    pb: Time,
}

struct Time {
    real_time: Option<TimeSpan>,
    game_time: Option<TimeSpan>,
}
```

We absolutely need to design this together with the splits i/o people and need to do this really soon as they are beginning to diverge too far.

glacials:
Cool! Some thoughts:

If the Splits.io exchange format is in good shape (or if we can get it there), we can get a lot of this stuff for free with something like [JSON Patch][1]. That way we won't need to maintain both a format for state and a protocol for events (the "events" would be defined by the state format). Plus clients won't need to know both how to read the state format and how to apply changes to it that come from events; they just constantly display their current state and blindly apply updates to it using a generic patching library.

Otherwise I would be very concerned about timers staying in sync if an event fails to process / implementation difference causes ripple effects in the run / etc.

[...]

PR #819, opened 2024-06-22
Implement the Server Protocol

CryZe:
This implements a new improved version of the server protocol. The following changes have been made:
- The protocol is based on JSON messages. This allows for example for more structured commands where it's easier to provide multiple arguments for a command and even have optional arguments.
- For each command, there is a corresponding response. It is either a `success` response with possibly the value that you requested, or an `error` response with an error `code`.
- On top of the responses you also get sent `event` messages that indicate changes to the timer. These can either be changes triggered via a command that you sent or by changes that happened through other sources, such as the user directly interacting with the timer or an auto splitter.

The protocol is still work in progress and we will evolve it into a protocol that fully allows synchronizing timers over the network. #260 

The event sink has now been renamed to command sink, because there is now a clear distinction between incoming commands and events that are the results of these commands.


CryZe:
Oh and because we now have specific error conditions for all the commands (split, undo_split, ...), there's a test for every possible error on every command. I think that should really help with robustness.

PR #823, opened 2024-07-08
Added runtime_get current_tick

Jujstme:
This adds a new tick field that increases by 1 at every call of the update function. This can be used for different purposes, eg. for caching data or to check (together with the current tick rate, albeit in an approximate way), the amount of time passed when executing code between iterations of the main autosplitting loop.


CryZe:
I'm not sure I like it. This adds an atomic increment (slower than a normal increment even) to every update tick, even if the auto splitter never even needs this functionality.

PR #856, opened 2024-12-28
Disable event tap on macOS before stopping run loop

### LINKED ISSUE #855: Hook does not uninstall cleanly on macOS
Try running this program:

```rust
use livesplit_hotkey::{ConsumePreference, Hook, Hotkey, KeyCode, Modifiers};
use std::sync::{Arc, Mutex};

fn main() {
    {
        let hook = Arc::new(Mutex::new(Some(
            Hook::with_consume_preference(ConsumePreference::MustConsume).unwrap(),
        )));
        let hook2 = Arc::clone(&hook);
        let mut hook = hook.lock().unwrap();
        let hook = hook.as_mut().unwrap();
        hook.register(
            Hotkey {
                modifiers: Modifiers::ALT,
                key_code: KeyCode::KeyV,
            },
            || println!("boo!"),
        )
        .unwrap();
        hook.register(
            Hotkey {
                modifiers: Modifiers::ALT,
                key_code: KeyCode::KeyB,
            },
            move || {
                println!("uninstalling");
                hook2.lock().unwrap().take();
            },
        )
        .unwrap();
    }
    std::thread::park();
}
```

Press alt+v any number of times to verify the hook is working. Verify that typing in the terminal causes text to show up, as normal.

Then press alt+b. "uninstalling" should appear.

Now try pressing some keys. The keys within about a second of the first press do not show up. After a second, keypresses start behaving normally again.

---

This sounds like macOS noticing an unresponsive event tap and removing the event tap, but I haven't verified this directly. I'm not sure why this is since the code isn't obviously wrong in any way, and appears to be handling reference counts correctly.

tmandry:
This shouldn't be necessary according to the documentation, but it fixes #855.

### DIFF
--- a/crates/livesplit-hotkey/src/macos/cg.rs
+++ b/crates/livesplit-hotkey/src/macos/cg.rs
@@ -220,5 +220,7 @@ extern "C" {
         userInfo: *mut c_void,
     ) -> MachPortRef;
 
+    pub fn CGEventTapEnable(tap: MachPortRef, enable: bool);
+
     pub fn CGEventGetIntegerValueField(event: EventRef, field: EventField) -> i64;
 }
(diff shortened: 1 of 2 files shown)


CryZe:
Looks good. I guess it can't hurt to disable it, even if the documentation doesn't say so.

PR #857, opened 2024-12-29
Bump hotkey to 0.8.0

tmandry:
Could you please publish a release of livesplit-hotkey?

This bumps to 0.8.0 because it includes at least one breaking change, in #757.


CryZe:
Oh sorry, I didn't know this was a PR, I only so the message in the mail. Anyways, I released 0.8.0.

PR #858, opened 2025-01-14
Fix issue of type i64 in Rust not properly getting converted to BigInt in TypeScript

kdevcse:
When compiling `livesplit-core` to WASM TypeScript from Rust, a compilation error is produced in the resulting `index.ts` file for the `wholeSeconds` function. The return type for this function is incorrectly set to `number` instead of `BigInt`, this PR corrects that problem by modifying the `wasm_bindgen.rs` file appropriately.

Admittedly, I don't know much about Rust so I could be handling this wrong, but I believe this is the correct fix and it seems to be working correctly for me. Let me know if I need to make any additional changes to get this working properly again 👍

### DIFF
--- a/capi/bind_gen/src/wasm_bindgen.rs
+++ b/capi/bind_gen/src/wasm_bindgen.rs
@@ -27,7 +27,7 @@ fn get_hl_type_without_null(ty: &Type) -> String {
                 "i8" => "number",
                 "i16" => "number",
                 "i32" => "number",
-                "i64" => "number",
+                "i64" => "BigInt",
                 "u8" => "number",
                 "u16" => "number",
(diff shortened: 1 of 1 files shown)


CryZe:
Thanks, I did notice that recent wasm-bindgen versions apparently changed the way number types are represented, but didn't yet get to fixing that, so thank you. Can you can it for U64 as well? Other than that, this looks good.

kdevcse:
> Thanks, I did notice that recent wasm-bindgen versions apparently changed the way number types are represented, but didn't yet get to fixing that, so thank you. Can you change it for U64 as well? Other than that, this looks good.

Done!

PR #867, opened 2025-03-25
Move Text Shadow Support to Rendering Backends

CryZe:
Instead of duplicating every label, it is much better for the rendering backends to handle the text shadow support by themselves. This also enables them to do it in a faster or more sophisticated way, such as using a proper text shadow blur.

## Web Backend

We use the `shadowColor`, `shadowOffsetX`, `shadowOffsetY`, and `shadowBlur` properties of the `CanvasRenderingContext2D` to draw the text shadows.

## Software Backend

We still basically just duplicate the labels. `tiny-skia` does not have a good way to add shadows or have any support for blurring.

## SVG Backend

We unfortunately also just duplicate the labels. The reason is that we tried to use a `feDropShadow` filter, but it seems to be cut off in Chrome and weirdly enough **segfaults** Firefox.

https://x.com/CryZe107/status/1903132316358082813

### DIFF
--- a/src/rendering/entity.rs
+++ b/src/rendering/entity.rs
@@ -6,8 +6,8 @@ use core::{
 use crate::settings::BackgroundImage;
 
 use super::{
-    resource::{Handle, LabelHandle},
     Background, FillShader, Rgba, Transform,
+    resource::{Handle, LabelHandle},
 };
 
 struct FxHasher(u64);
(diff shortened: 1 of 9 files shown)


Skajdrowski:
The shadow offset for spliits & key/value seem to look too low, while timer shadow is now too high.
It would be nice, if there was a setting to adjust offsets individually. I wasn't sure where to implement them

CryZe:
You are right, the behavior slightly changed, but is closer to how the original LiveSplit did it. We can definitely add more settings.

PR #875, opened 2025-06-14
Add Current Segment component

thirstyice:
Basically does what the previous segment component does, but for the current segment.
In other words, it does what the previous segment component does when it's in Live Segment mode, but all the time.

> The Current Segment Component is a component that shows how much time will be saved or lost during the current segment based on the chosen comparison. It displays the difference between the current segment time and the chosen comparison segment time.

### DIFF
--- a/capi/src/current_segment_component.rs
+++ b/capi/src/current_segment_component.rs
@@ -1,0 +1,59 @@
+//! The Current Segment Component is a component that shows how much time will
+//! be saved or lost during the current segment based on the chosen comparison.
+//! It displays the difference between the current segment time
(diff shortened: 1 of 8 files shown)


wooferzfg:
Why couldn't this instead be added as a setting to the Delta component?

thirstyice:
Two reasons I did it this way:
1. Adding it as a setting to the Delta component would require a fair bit of re-writing of the Delta component (from what I can tell, the Delta component is optimized for dealing with full-run deltas, less so for individual segments)
2. I'm not sure how to implement a new setting yet

PR #884, opened 2025-08-31
Add timer_current_split_index and timer_segment_splitted

AlexKnauth:
Adds timer_current_split_index and timer_segment_splitted, which allow autosplitters to detect split/skip/undo events, even if they were done manually by the user.

I've tested it with a demo autosplitter that sets custom variables to display the events it detects, and recorded that here: https://www.youtube.com/watch?v=l1kdwRjxT9Y

This PR is an alternative to https://github.com/LiveSplit/livesplit-core/pull/882, implementing similar functionality by a different method. Using https://github.com/LiveSplit/livesplit-core/pull/882 's timer_current_attempt_segments_splitted requires using alloc, but using this PR's timer_current_split_index and timer_segment_splitted does not.

### DIFF
--- a/src/auto_splitting/mod.rs
+++ b/src/auto_splitting/mod.rs
@@ -105,9 +105,12 @@
 //!     /// when the attempt is finished, but has not been reset.
 //!     /// So you need to be careful when using this value for indexing.
 //!     /// Same index does not imply same split on undo and then split.
-//!     pub safe fn timer_current_split_index() -> i32;
-//!     /// Whether the segment given by `idx` was splitted this attempt.
-//!     pub safe fn timer_segment_splitted(idx: i32) -> bool;
+//!     pub safe fn timer_current_split_index() -> i64;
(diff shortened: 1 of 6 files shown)


src/auto_splitting/mod.rs
```diff
@@ -800,13 +803,19 @@ impl<E: event::CommandSink + TimerQuery + Send> AutoSplitTimer for Timer<E> {
...
-    fn segment_splitted(&self, idx: usize) -> bool {
-        self.0
-            .get_timer()
-            .run()
-            .segments()
-            .get(idx)
-            .is_some_and(|segment| segment.split_time().real_time.is_some())
+    fn segment_splitted(&self, idx: usize) -> Option<bool> {
+        let t = self.0.get_timer();
+        if !(idx < t.current_split_index()?) {
```
CryZe, line 808: Doesn't the .get(idx) already do that check?

PR #902, opened 2026-02-25
WIP: Start on LiveSplit Server Protocol

AlexKnauth:
> I wonder if we could somehow unify both protocols (as in the new one is v2 or so and it gets negotiated)

> would it work like... in livesplit-core server protocol, having it try to parse it as JSON, but when that fails, fall back on a different parser that recognizes the same protocol as normal LiveSplit?
> and in normal LiveSplit, something similar, recognizing commands in either format?

> possibly

A WIP incomplete translation from [`LiveSplit/src/LiveSplit.Core/Server/CommandServer.cs`](https://github.com/LiveSplit/LiveSplit/blob/master/src/LiveSplit.Core/Server/CommandServer.cs).

I have only tested the `getsplitindex` command so far, the rest of the commands are currently untested.

Missing these commands entirely:
- `addloadingtimes`
- `alwayspausegametime`
- `setsplitname` / `setcurrentsplitname`

### DIFF
--- a/src/networking/server_protocol.rs
+++ b/src/networking/server_protocol.rs
@@ -64,14 +64,19 @@ pub async fn handle_command<S: event::CommandSink + event::TimerQuery>(
     command: &str,
     command_sink: &S,
 ) -> String {
-    let response = match serde_json::from_str::<Command>(command) {
-        Ok(command) => command.handle(command_sink).await.into(),
-        Err(e) => CommandResult::Error(Error::InvalidCommand {
-            message: e.to_string(),
-        }),
-    };
+    if serde_json::from_str::<serde_json::Value>(command).is_ok() {
+        let response = match serde_json::from_str::<Command>(command) {
+            Ok(command) => command.handle(command_sink).await.into(),
+            Err(e) => CommandResult::Error(Error::InvalidCommand {
+                message: e.to_string(),
+            }),
+        };
 
-    serde_json::to_string(&response).unwrap()
+        serde_json::to_string(&response).unwrap()
+    } else {
+        // fall back on normal LiveSplit's server protocol
+        handle_livesplit_command(command, command_sink).await
+    }
 }
 
 /// Encodes an event that happened to be sent.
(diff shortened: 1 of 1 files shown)


AlexKnauth:
See also https://github.com/LiveSplit/livesplit-core/pull/955

devek1:
I think the condition `serde_json::from_str::<serde_json::Value>(command).is_ok()` should probably be replaced with `command.trim_ascii_start().starts_with('{')`, which should be a more performant check than a full JSON parse, and a more complete coverage of inputs that are **trying** to use the new protocol and always giving actual error messages for those, rather than treating invalid JSON as old protocol and not reacting to it at all. (I believe JSON that starts with an array is irrelevant to the LSO protocol but it could also be accounted for by changing `'{'` to `&['{','[']`)

the `trim_ascii_start` part is probably not really necessary since I don't see a reason for anything using the LSO protocol to include leading whitespaces, but they are technically valid in JSON.

AlexKnauth:
Okay, I've incorporated that as `command.trim_start().starts_with(&['{', '['])`.

In the time between when I first translated this and now, LiveSplit has added several new commands. I should translate those as well before I think of marking this as ready for review.

The other main thing is testing. As far as I know, there are no integration tests for LiveSplit's server protocol. If there were, and if those integration tests also passed for this PR on a version of LiveSplit One that incorporates it such as LiveSplit One Druid, I would feel more confident in the readiness of this PR.

devek1:
Following commands are working for me: (using a livesplit-one-druid build as a LiveSplit server, and a Node.js net.Socket in an RPG Maker MV game as the client)
```
start | starttimer
reset
startorsplit
split
skipsplit
unsplit | undosplit
pause
resume
pausegametime
unpausegametime
setloadingtimes
setgametime
setcomparison COMPARISON
setcustomvariable ["key","val"] (though I'm not sure if that format for the arguments is how it's meant to work)
getcustomvariableval KEY
getcurrenttime
getcurrentrealtime
getcurrentgametime
getlastsplittime
gettimerphase (returns a value but IDK if the names are accurate to what the legacy protocol expects) 
getsplitindex (value starts from 0, returns -1 if timer not started)
getdelta
getcurrentsplitname
getlastsplitname
getcurrentsplittime
getfinaltime
ping
getattemptcount
getcompletedcount
```

however, I've noticed that `undoallpauses` is outright missing

AlexKnauth:
I've just added a bunch more commands, including `undoallpauses`
(Edit: though I haven't updated LiveSplit One Druid to point to this new commit yet)

PR #925, opened 2026-06-02
Add "Merged Timeloss" comparison

Mian-K:
For this comparison, all timeloss is merged into the first split, with all other splits set to their best segment time.

My reason is just that I want to know how much time I can still lose in total, rather then see it split up over all segments.

### DIFF
--- a/src/comparison/merged_timeloss.rs
+++ b/src/comparison/merged_timeloss.rs
@@ -1,0 +1,53 @@
+//! Defines the Comparison Generator for calculating the Best Segments of a
+//! [`Run`](crate::Run).
+
+use super::ComparisonGenerator;
+use crate::{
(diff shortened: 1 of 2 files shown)


wooferzfg:
Why not use the "Possible Time Save" component with the "Total Possible Time Save" option?

Mian-K:
I was fully unaware of the existence of this component, though putting it into a layout and just seeing how it works feels unhelpful.
Set to "Total Possible Time Save", it shows how much future segments are behind their respective best, but ignore prior segments being behind or ahead. The current segment gets live updated once you go slower then the best for the segment, and fully stops once you are behind the segment time of the comparison.

For other components that would be closer to what I'd want there is "Current Pace" with the Comparison set to "Best Segments". It's showing the theoretical finish time rather then showing the Delta compared to PBs finish time, though that would at least be something.

PR #927, opened 2026-06-13
fix(hotkey): support remapped Caps Lock (hyper key) on macOS

### LINKED ISSUE #895: Hotkey remapped as caps lock in raycast doesn't trigger
livesplit-hotkey v0.8.0, macOS Tahoe 26.2

I have Caps Lock bound to my hyper key, <kbd>Ctrl + Opt + Cmd</kbd>, using the Raycast hyper key setting: https://manual.raycast.com/hyperkey

I am able to get the callback to trigger when pressing those keys manually, but not when pressing the Caps Lock key that is rebound to those keys via raycast. This is specific to the caps lock key, if I rebind another key to <kbd>Ctrl + Opt + Cmd</kbd> then it works fine.

Here's is the reproduction code:

```rs
use livesplit_hotkey::{Hook, Hotkey, KeyCode};

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let hook = Hook::new()?;
    let hyper_key_a = Hotkey {
        key_code: KeyCode::KeyA,
        modifiers: livesplit_hotkey::Modifiers::META
            | livesplit_hotkey::Modifiers::ALT
            | livesplit_hotkey::Modifiers::CONTROL,
    };
    hook.register(hyper_key_a, || {
        println!("Hyper + A was pressed!");
    })?;
    loop {
        std::thread::sleep(std::time::Duration::from_secs(1));
    }
}
```

And of course you'll need to have raycast running with kyperkey bound to caps lock.

I can replicate this with all 4 consume settings options.

hajiboy95:
Check both device-independent and device-dependent modifier flags when processing events. This fixes an issue where virtual keys/remappers (such as Raycast or Karabiner-Elements Hyper Key) only set the device-dependent flags, preventing the hotkey from triggering.

Closes #895

### DIFF
--- a/crates/livesplit-hotkey/src/macos/mod.rs
+++ b/crates/livesplit-hotkey/src/macos/mod.rs
@@ -456,33 +456,25 @@ unsafe extern "C" fn callback(
     let modifier_flags = unsafe { CGEventGetFlags(event) };
     let mut modifiers = Modifiers::empty();
 
-    if (modifier_flags.contains(EventFlags::SHIFT)
-        || modifier_flags.contains(EventFlags::LEFT_SHIFT)
-        || modifier_flags.contains(EventFlags::RIGHT_SHIFT))
+    if modifier_flags.intersects(EventFlags::SHIFT | EventFlags::LEFT_SHIFT | EventFlags::RIGHT_SHIFT)
         && !matches!(key_code, KeyCode::ShiftLeft | KeyCode::ShiftRight)
     {
         modifiers.insert(Modifiers::SHIFT);
     }
 
-    if (modifier_flags.contains(EventFlags::CONTROL)
-        || modifier_flags.contains(EventFlags::LEFT_CONTROL)
-        || modifier_flags.contains(EventFlags::RIGHT_CONTROL))
+    if modifier_flags.intersects(EventFlags::CONTROL | EventFlags::LEFT_CONTROL | EventFlags::RIGHT_CONTROL)
         && !matches!(key_code, KeyCode::ControlLeft | KeyCode::ControlRight)
     {
         modifiers.insert(Modifiers::CONTROL);
     }
 
-    if (modifier_flags.contains(EventFlags::OPTION)
-        || modifier_flags.contains(EventFlags::LEFT_OPTION)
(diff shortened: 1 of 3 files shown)


tom-sherman:
Tested this and looks great, solves the bug. @CryZe can we get this landed? 🙇

crates/livesplit-hotkey/src/macos/mod.rs
```diff
@@ -456,25 +456,33 @@ unsafe extern "C" fn callback(
     let modifier_flags = unsafe { CGEventGetFlags(event) };
     let mut modifiers = Modifiers::empty();
 
-    if modifier_flags.contains(EventFlags::SHIFT)
+    if (modifier_flags.contains(EventFlags::SHIFT)
+        || modifier_flags.contains(EventFlags::LEFT_SHIFT)
+        || modifier_flags.contains(EventFlags::RIGHT_SHIFT))
```
CryZe: I believe it would be better to write this as `.intersects(EventFlags::SHIFT | EventFlags::LEFT_SHIFT | EventFlags::RIGHT_SHIFT)` or so (I haven't check the exact API).

hajiboy95:
@CryZe adressed. Thanks

CryZe:
The build error is related to this (independent of your changes): https://blog.rust-lang.org/2026/04/04/changes-to-webassembly-targets-and-handling-undefined-symbols/

If you want to fix it, feel free to (either as part of this PR or another). Otherwise I can fix it later today.

hajiboy95:
@CryZe I've commit a possible fix. please let me know if that's what you had in mind! Otherwise I'd just pass it forward to you and could revert this change.

CryZe:
Oh sorry, I should've been clearer (and possibly the blog post too), allow-undefined is not the correct solution. The solution is `#[link(wasm_import_module = "env")]` on the extern "C" { ... } block.

I believe it should be these two extern "C" blocks: https://github.com/LiveSplit/livesplit-core/blob/master/src/platform/wasm/unknown/time.rs#L11-L28

hajiboy95:
@CryZe here we go :D thanks for the clarification

PR #934, opened 2026-07-06
Add Native Subsplit Support

### LINKED ISSUE #36: Implement Sub Splits into Splits Component

CryZe:
This introduces native subsplit support to `livesplit-core` by representing subsplits as first-class segment groups over the existing flat `Run` segment list. The flat list remains the canonical structure for timing, comparisons, history, auto splitting, and editor indices, while `SegmentGroup` / `SegmentGroups` provide a validated alternate view for display, editing, and serialization. Groups are one-level contiguous ranges, with the final `Segment` in the range acting as the major split, and group names and icons can either be explicit or inherit from that major split.

The `LiveSplit` XML format now stores segment groups directly after the segment list in `<SegmentGroups>`. Explicit group metadata is preferred whenever present, while older files without native segment groups are upgraded from the legacy dash-based subsplit convention. The legacy importer strips one leading dash from subsplit names, groups contiguous subsplits with the following major split, and supports the `{Group Name}` prefix syntax for giving a group a display name separate from the major split name. New-format files no longer interpret leading dashes as subsplit syntax.

The `Splits` component now supports these groups directly instead of requiring a separate subsplits component. `SubsplitDisplayMode` can show the flat list, show all groups expanded, or show a hierarchical `CurrentGroupExpanded` view where only the current or manually scrolled-to group is expanded. Closed groups summarize the major split and show the appropriate columns for the whole group, while expanded groups render a header and indented segment rows. When the timer is `NotRunning`, the hierarchical cursor starts before the first split, and when the run has `Ended`, it starts after the final split, so groups are collapsed by default in those states but can still be opened by manually scrolling into them.

The `RunEditor` now exposes segment groups as editable structure rather than name syntax. It can create, remove, rename, and preserve groups through segment movement, insertion, deletion, and selection changes. Group headers are represented in the editor state, group names can be left automatic via empty strings, and group icons can be set explicitly or inherited from the major split icon. The `C API` and generated bindings expose the new editor operations and state so frontends can edit groups directly while keeping flat segment indices stable.


capi/bind_gen/src/typescript.ts
```diff
@@ -986,6 +993,16 @@ export interface RunEditorButtonsJson {
...
+    /**
+     * Describes whether the currently selected segments can be turned into a
+     * segment group.
+     */
+    can_create_segment_group: boolean,
+    /**
+     * Describes whether the currently selected segments are exactly one
+     * segment group that can be removed.
+     */
+    can_remove_segment_group: boolean,
```
CryZe: What about removing multiple groups?

capi/bind_gen/src/typescript.ts
```diff
@@ -1012,6 +1029,30 @@ export interface RunEditorRowJson {
...
+     * display name of the group.
+     */
+    name: string | null,
+    /**
+     * The group display icon. This falls back to the major split icon if no
+     * explicit group icon is set.
+     */
+    icon: ImageId,
+    /** Whether the group icon is explicitly set instead of inherited. */
+    has_explicit_icon: boolean,
```
CryZe, line 1087: This feels redundant, we don't do this for the segment icons either.

capi/bind_gen/src/typescript.ts
```diff
@@ -1012,6 +1029,30 @@ export interface RunEditorRowJson {
...
+}
+
+/** Describes a segment's role in a segment group. */
+export interface RunEditorSegmentGroupStateJson {
+    /** The index of the group this segment belongs to, if any. */
+    group_index: number | null,
+    /** Whether this segment is a subsplit inside the group. */
+    is_subsplit: boolean,
+    /** Whether this segment is the major split ending the group. */
+    is_major_split: boolean,
```
CryZe: Do we need is_major_split? I feel like this entire way of communicating the groups is weird in how they are attached to individual splits.

PR #946, opened 2026-08-05
Defer WASI startup until the first update

CryZe:
WebAssembly start functions currently run during instantiation, before callers can obtain an interrupt handle, so a hanging initializer can block the runtime indefinitely. Defer initialization until the first update so startup code is covered by the normal interruption mechanism.


AlexKnauth:
If an autosplitter uses WASI with `std` and tries to run `File::open` and `File::read_to_end` on the first tick, before it ever calls `next_tick().await;`, what will happen? Will `File::open` return an `Err` result value where it had previously returned an `Ok` result?

PR #950, opened 2026-08-16
Add splits path getter for autosplitters

caeyo:
Allows autosplitters to access the WASI filesystem path of the current splits file (if one is loaded).

Related PRs:
- https://github.com/LiveSplit/asr/pull/138
- https://github.com/LiveSplit/LiveSplit.AutoSplittingRuntime/pull/28

### DIFF
--- a/crates/livesplit-auto-splitting/README.md
+++ b/crates/livesplit-auto-splitting/README.md
@@ -249,6 +249,13 @@ unsafe extern "C" {
     /// guaranteed to be valid UTF-8 and is not nul-terminated.
     /// Example values: `windows`, `linux`, `macos`
     pub fn runtime_get_os(buf_ptr: *mut u8, buf_len_ptr: *mut usize) -> bool;
+    /// Stores the WASI path of the currently loaded splits file in the buffer
+    /// given. Returns `false` if there is no splits path or the buffer is too
+    /// small. After this call, no matter whether it was successful or not, the
+    /// `buf_len_ptr` will be set to the required buffer size, or `0` if there
+    /// is no splits path. The path is guaranteed to be valid UTF-8 and is not
+    /// nul-terminated.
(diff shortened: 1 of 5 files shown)


CryZe:
I think once again, this one needs some justification as well. Of course, it probably makes some amount of sense, but I want to hear what you want to use this for and maybe there's some better API we can add where you can just query more information about the splits in general.

AlexKnauth:
I can't speak for caeyo, but the main reason I would have reached for something like `runtime_get_splits_path` is for an ASR auto splitter that's meant to be compatible with the settings of an existing dedicated Component auto splitter: so that splits files made with or for the previous auto splitter don't have their settings erased when you open them with the new ASR auto splitter. This is a problem that could be solved multiple ways, with my preferred solution being a way for ASR to query the Legacy Raw XML of just the `AutoSplitterSettings` element from the run.

Another reason could be for supporting a version of a "settings per split" pattern where the number of segments and the segment names can be displayed to the user in the auto splitter settings editor, so that they know exactly which setting corresponds with which segment while they're editing it... but full support of this would require other changes such as allowing ASR to remove widgets in the event that the user has removed segments, and rename title widgets in the event that the user has renamed segments, in addition to a way of getting the number and names of the segments... or just a full settings per split implementation some other way which would be cool but seems far off.

caeyo:
Yeah Alex hit it on the head here - this was my way of opening up migrating an old Component’s settings in an ASR splitter. 

I had a think about proposing a proper API for migration, but as far as I can tell it’s impossible to distinguish between the ASR settings format and an old ASL script’s settings format, so it seemed necessary to add something that gives the splitter access to the raw XML to let it decide.

I’ve only ever really been in development environments where it’s preferred if additions can serve use cases beyond those currently identified, and as such I went with exposing the splits path over exposing the legacy XML since there’s more a splitter can do with the former - correct me if I’m wrong but my read of the philosophy in this project now is that features should serve narrow use cases, so I don’t mind if exposing the legacy XML is the preferred path.

PR #951, opened 2026-08-16
Allow autosplitters to set timer's timing method

caeyo:
Related PRs:
- https://github.com/LiveSplit/LiveSplit.AutoSplittingRuntime/pull/29
- https://github.com/LiveSplit/asr/pull/139

### DIFF
--- a/crates/livesplit-auto-splitting/src/lib.rs
+++ b/crates/livesplit-auto-splitting/src/lib.rs
@@ -140,6 +140,8 @@
 //!     /// Resumes the game time. This does not resume the timer, only the
 //!     /// automatic flow of time for the game time.
 //!     pub safe fn timer_resume_game_time();
+//!     /// Sets whether the timer compares against real time or game time.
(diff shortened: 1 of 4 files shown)


AlexKnauth:
This should not be added as-is. An auto-splitter should not be able to override a user's chosen timing method.

Maybe it can be a "willUseGameTime" hint function, or maybe it can be done with better metadata handling, possibly with SRC or `LiveSplit.AutoSplitters.xml` metadata, without adding anything to the `asr` API at all, but either way, there should be a way for users to either set a setting to prevent the auto-splitter's timing method choice from overriding their own, or for LiveSplit to prompt the user and ask permission before setting it.

caeyo:
As I've stated in the Discord, I'm happy to provide an alternative implementation, but I need some sort of proposal sanctioned by a maintainer because plenty of people raised plenty of issues – I like a simple `ask_to_set_timing_method` function that hosts can handle however they like

PR #952, opened 2026-08-16
Detailed Timer timing method override fix to apply to segment timer

mxous:
I override my timing method to always be Game Time but found that the override was not being applied to the segment timer but it was being applied to the main timer.

Fix just has the override apply to the segment timer as well.

Included a test if you're into that.


CryZe:
Oh wow good find. Thank you for fixing this.

PR #955, opened 2026-08-26
Partial support for legacy server protocol commands

devek1:
Adds support for the following commands from the legacy LiveSplit server protocol:
```
start/starttimer
split
startorsplit
reset
unsplit/undosplit
skipsplit
pause
resume
undoallpauses
setcomparison NAME
switchto realtime/gametime
initgametime (doesn't seem to have been a real command in the legacy protocol but I'm familiar with some clients that try it)
setgametime TIME
pausegametime
unpausegametime
setloadingtimes TIME
```
This does NOT currently cover all commands that exist in both protocols. In particular, none of the commands that expect a response from LiveSplit are supported as they would require using a different response format when responding to their legacy-protocol versions, and support for setcustomvariable is not yet included.

### DIFF
--- a/src/networking/server_protocol.rs
+++ b/src/networking/server_protocol.rs
@@ -59,6 +59,41 @@ use crate::{
     timing::formatter::{self, ASCII_MINUS, TimeFormatter},
 };
 
+/// Handles commands from a subset of the original LiveSplit's IPC/TCP/WebSocket server protocol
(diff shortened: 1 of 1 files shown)


AlexKnauth:
See also https://github.com/LiveSplit/livesplit-core/pull/902

devek1:
looks like #902 is a lot more complete and the way it's implemented probably makes more sense

